PR #3, opened 2016-09-06
fix libraryTarget

yale:
this allows you to do this:

```js
import Plasma from '@wework-dev/plasma';
```

instead of having to do this:

```js
import { Plasma } from '@wework-dev/plasma';
```


nickstamas:
we should probably remove index.js from the repo

PR #75, opened 2017-03-29
add disabled background color for table row

rfarine:


nickstamas:
@rfarine can you add screenshots?

nickstamas:
Just a note, in the future I think we should bump the version as a separate thing from PR.

PR #83, opened 2017-04-24
add search component with option to be clearable

rfarine:
Default
screen shot 2017-04-24 at 3 49 46 pm

Focused and with Text
screen shot 2017-04-24 at 3 49 56 pm

With Instructions
screen shot 2017-04-24 at 3 50 03 pm

With Clearable Icon
screen shot 2017-04-24 at 3 50 11 pm

Disabled
screen shot 2017-04-24 at 3 50 17 pm


nickstamas:
@rfarine This looks good. One question, do you think there's enough of a difference between a Search component and a TextInput component to justify a new component vs moving the logic into TextInput? I'm fine with either, and I'm not sure how I feel about it either way. What do you think?

rfarine:
@nickstamas There's a lot of other features that the Search component is going to need.. like a dropdown for autocomplete, etc... so I created this for the future of the  component.

nickstamas:
@rae True, I guess it starts to overload a TextInput with a lot of stuff going forward. Thanks for adding this :)

rfarine:
@nickstamas No worries! :D

nickstamas:
👍

PR #85, opened 2017-05-03
Add Tabs component

kangax:
I made this pretty minimal for now — just a way to pass an array of strings and a single callback for all of them that's being passed an index of a selected tab.

Added support for the keyboard as well since it's not too complicated although not sure how important that is at the moment. It'd be nice to add ARIA roles as well but that'd probably be overkill right now.

### DIFF
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -63,6 +63,6 @@ Tabs.propTypes = {
   onClick: PropTypes.func,
 };
 
-Tabs.displayName = 'Tab';
+Tabs.displayName = 'Tabs';
 
 export default Base(Tabs);
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -1,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Base from '../Base.jsx';
+import style from './style.scss';
+import cn from 'classnames';
+
+class Tabs extends React.Component {
+  constructor() {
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,68 @@
...
+    this.state = {
+      selectedIndex: 0
+    };
+  }
+  onClick(index) {
+    this.setState({
+      selectedIndex: index
+    });
+    if (this.props.onClick) {
+      this.props.onClick(index);
```
yale: i'm wondering if we can somehow give a `key` to each tab, so that what's passed to the handler is like, 'my_tab' and 'another_tab' instead of `0` or `1`. that would be less fragile if we were to add new tabs or reorder existing ones.

yale:
the style & component in general look good, however see my comment about using some sort of `key` in the onClick handler.

yale:
whoops didn't mean to close, sorry!

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Base from '../Base.jsx';
+import style from './style.scss';
+import cn from 'classnames';
+import _ from 'lodash';
+
+class Tabs extends React.Component {
+  constructor(props) {
+    super(props);
+
+    let selectedIndex = _.findIndex(props.items, item => item.selected);
```
yale: minor: maybe the lodash shorthand exists for this: `_.findIndex(props.items, 'selected')`

PR #86, opened 2017-05-03
Fix all lint errors & setup lint task

kangax:


kangax:
@yale we gotta figure out how to avoid these yarn issues so that we can iterate and merge things fast. Let me see if we can upgrade single package with yarn...

PR #87, opened 2017-05-12
Fixed the alignment of the check within the checkbox

deenaedw:


src/components/Checkbox/style.scss
```diff
@@ -62,8 +62,6 @@
     display: block;
     content: '';
     background: none;
-    width: 12px;
```
nickstamas: oh nice!

PR #95, opened 2017-06-21
MT-2571 - NumberInput Fixed Height

williamtan37:


src/components/NumberInput/NumberInput.jsx
```diff
@@ -63,6 +66,7 @@ class NumberInput extends React.Component {
 }
 
 NumberInput.propTypes = {
+  wrapperClass: PropTypes.object,
```
yale: are you sure this is an object and not a string?

rfarine:
@yale https://jira.we.co/browse/MT-2571

PR #100, opened 2017-06-23
Fix Unknown prop `description` on <input> tag console warning

aaroncouture:
This fixes a prop warning on the RadioButton component.. description is not a valid prop for input.

### DIFF
--- a/src/components/RadioButton/RadioButton.jsx
+++ b/src/components/RadioButton/RadioButton.jsx
@@ -21,7 +21,6 @@ class RadioButton extends React.Component {
             type="radio"
             id={id}
             name={this.props.name}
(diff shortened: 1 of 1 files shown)


kangax:
Should it be removed from `RadioButton.propTypes` as well?

kangax:
I also wonder what was the purpose of this in the first place. Should we use "title" attribute instead? To show tooltip with a hint.

aaroncouture:
@kangax the description prop is being used in the component (it is being rendered into a span). Not sure why it was also being passed into the input tag as well (since that isnt a supported attribute). In short, that propType is still needed.

kangax:
Ah, I see. Makes sense then.

PR #101, opened 2017-06-24
Side nav redesign

nickstamas:


nickstamas:
This is a major cleanup for Plasma. Biggest things is reducing bundle size down to < 300KB.

- Gets rids of jQuery dependency!
- Selectively pulls functions from lodash instead of dependency on entire library!
- Adds Plasma@ prefix to displayName for components so it's easier to identify Plasma components!
- Eliminates `Base` wrapper component and rely on `getDataAttrs` method to pass through `data` props!
- Adds new SideNav component. Keeping the other SideNav component around during transition.

screen shot 2017-06-24 at 2 31 07 pm

kangax:
Wow, this is epic! ❤️

I'll look through in a bit but just wanted to mention that there's a missing exclamation mark in the last bullet point.

src/components/Modal/Modal.jsx
```diff
@@ -28,15 +54,21 @@ class Modal extends React.Component {
...
 }
 
-Modal.defaultProps = {};
+Modal.defaultProps = {
+  minWidth: 'auto',
+};
 
 Modal.propTypes = {
+  title: PropTypes.string,
+  actions: PropTypes.array,
```
aaroncouture: actions: PropTypes.arrayOf(PropTypes.element)

src/components/Icon/style.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../styles/base';
+
+.wrapper {
+  fill: red;
```
kangax: red filling by default?

src/components/Label/Label.jsx
```diff
@@ -15,30 +14,36 @@ const Label = (
...
+    data,
+    children,
   }
 ) => {
   const labelStyle = cx(style.label, {
     [style.labelSecondary]: type === types.SECONDARY,
     [style.labelDisabled]: type === types.DISABLED,
   });
-
+  console.log(children);
```
kangax: /gif kill it with fire

src/components/Modal/Modal.jsx
```diff
@@ -1,25 +1,51 @@
...
+                  <div className={style.top}>
+                    <Header h3>
+                      { title }
+                    </Header>
+                  </div>
+                }
+                {children}
+                { actions &&
+                  <div className={style.pageActions}>
+                    { actions.map((action) => action)}
```
kangax: why? this is no-op

src/components/SideNavBar/OldSideNavBar.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import style from './oldStyle.scss';
+
+class SideNavBar extends Component {
```
kangax: should this be a stateless component then? I think those are supposed to be more performant eventually (in near future versions of React)

src/components/SideNavBar/SideNavBar.jsx
```diff
@@ -1,22 +1,69 @@
...
 class SideNavBar extends Component {
+
+  constructor() {
+    super();
+    this.renderItems = this.renderItems.bind(this);
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(result) {
+    console.log(result.id);
```
kangax: ditto

src/components/SideNavBar/SideNavBar.jsx
```diff
@@ -1,22 +1,69 @@
...
+import getDataAttrs from '../../getDataAttrs';
+import SideNavBarItemGroup from './SideNavBarItemGroup';
+import SideNavBarTop from './SideNavBarTop';
 import style from './style.scss';
 
 class SideNavBar extends Component {
+
+  constructor() {
+    super();
+    this.renderItems = this.renderItems.bind(this);
```
kangax: just define methods as arrow functions below and drop all this binding

src/components/SideNavBar/SideNavBar.jsx
```diff
@@ -1,22 +1,69 @@
...
+          id={item.id}
+          icon={item.icon}
+          label={item.label}
+          onClick={this.handleClick}
+          items={item.items}
+          selectedId={this.props.selectedId}
+        />
+      );
+    });
+    return items;
```
kangax: seems like `return this.props.items.map()` would be perfect here

src/components/SideNavBar/SideNavBarItem.jsx
```diff
@@ -1,59 +1,69 @@
...
-        src={icon}
-        alt={label}
-      />
-    );
+  constructor() {
+    super();
+    this._onClick = this._onClick.bind(this);
+  }
+
+  _onClick() {
```
kangax: also arrow function & ditch binding

README.md
```diff
@@ -4,26 +4,59 @@ Plasma is a design system for creating sensible, modern interfaces.
...
+1. Follow steps above "Running Storybook"
+2. Create a new dev branch: `git checkout -b mybranch`
+3. Add / update as necessary
+4. Add stories to the `stories/` directory as you add / update components!
+
+## Creating a Pull Request / Publishing
+
+### Creating a Pull Request
+
+1. When you're ready, pushyour  branch to Github and create a pull request.
```
yale: `push your`

src/components/SideNavBar/SideNavBarItemGroup.jsx
```diff
@@ -0,0 +1,117 @@
...
+import { forEach, find } from 'lodash';
+import SideNavBarItem from './SideNavBarItem';
+import style from './style.scss';
+
+class SideNavBarItemGroup extends Component {
+
+  constructor() {
+    super();
+    this.handleGroupClick = this.handleGroupClick.bind(this);
+    this.handleItemClick = this.handleItemClick.bind(this);
```
kangax: ditto

package.json
```diff
@@ -55,21 +57,22 @@
...
     "babel-runtime": "^6.9.2",
     "classnames": "^2.2.5",
-    "jquery": "^3.1.0",
     "lodash": "^4.14.1",
     "prop-types": "^15.5.7",
     "proxy-middleware": "^0.15.0",
     "react-datepicker": "^0.46.0",
     "react-input-autosize": "^1.1.0",
-    "react-select": "^0.9.1"
+    "react-select": "^1.0.0-rc.5"
```
yale: careful with this!

src/components/SideNavBar/SideNavBarItemGroup.jsx
```diff
@@ -0,0 +1,117 @@
...
+    this.props.onClick({ id: itemId });
+  }
+
+  renderIconAndLabel() {
+    const { icon, iconSize, iconStyle, label, items } = this.props;
+    if (!find(items, (item) => !item.hidden)) {
+      return null;
+    }
+
+    const renderIcon = icon && iconSize && (
```
kangax: minor: not to sound pedantic, but renderIcon sounds like a method name and kinda throws you off when reading code; can we do smth like `iconToRender`?

src/components/Icon/Icon.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+import getDataAttrs from '../../getDataAttrs';
+import style from './style.scss';
+
+const Icon = ({ icon, color, data }) => {
+  return (
+    <div
+      className={style.wrapper}
+      style={{ fill: color }}
+      dangerouslySetInnerHTML={{ __html: icon }}
```
yale: why dangerously set? why not just render the icon as a child?

src/components/Label/Label.jsx
yale: remove

src/components/Modal/Modal.jsx
```diff
@@ -1,25 +1,51 @@
...
     let comp = null;
+    const {
+      title,
+      actions,
+      data,
+      children,
+      onDismiss,
+      minWidth,
+    } = this.props;
     if (this.props.visible) {
```
yale: why not return early instead of setting a variable?

```js
if (!this.props.visible) {
  return null;
}

return <div ...
```

src/components/layout/FixedLeft/FixedLeft.jsx
```diff
@@ -85,8 +98,9 @@ FixedLeft.propTypes = {
   stickAt: PropTypes.number,
   contentStyle: PropTypes.object,
   fixedStyle: PropTypes.object,
+  data: PropTypes.object,
 };
 
-FixedLeft.displayName = '!FixedLeft';
+FixedLeft.displayName = '!Plasma@FixedLeft';
```
kangax: what does "!" mean?

src/components/layout/FixedRight/FixedRight.jsx
```diff
@@ -1,14 +1,25 @@
...
-import Base from '../../Base.jsx';
+import getDataAttrs from '../../../getDataAttrs';
 import style from './style.scss';
 
 /**
   * The fixed right page is great.
  */
 
+/* eslint-disable */
+const getOffsetTop = (elem) => {
```
kangax: should we share it via a lib or something?

src/components/layout/Page/Page.jsx
```diff
@@ -0,0 +1,61 @@
...
+          { (title || actions) &&
+            <div className={style.top}>
+              { title &&
+                <Header h2 noMargin>
+                  { title }
+                </Header>
+              }
+              { actions &&
+                <div className={style.pageActions}>
+                  { actions.map((action) => action)}
```
kangax: get rid of map also?

src/components/SideNavBar/SideNavBar.jsx
yale: no need to do this if use enable `transform-class-properties` in babel, like [this](https://github.com/WeConnect/spacestation-v2/blob/master/.babelrc#L6)

```jsx
class SideNavBar extends Component {
  // this will automatically be bound to the component instance
  renderItems = () => {
    ...
  }
  ...
}

File: src/components/SideNavBar/SideNavBar.jsx [hunk 19]
```diff
@@ -1,22 +1,69 @@
...
+  }
+
+  handleClick(result) {
+    console.log(result.id);
+    this.props.onChange && this.props.onChange({ id: result.id });
+  }
+
+  renderItems() {
+    const items = [];
+    forEach(this.props.items, (item) => {
```
yale:
```jsx
return this.props.items.map(item => (<SideNavBarItemGroup ... />));
```

src/components/SideNavBar/SideNavBarItem.jsx
```diff
@@ -1,59 +1,69 @@
...
-          height: iconSize,
-          ...iconStyle,
-        }}
-        src={icon}
-        alt={label}
-      />
-    );
+  constructor() {
+    super();
+    this._onClick = this._onClick.bind(this);
```
yale: see comment about `transform-class-properties`

src/components/SideNavBar/SideNavBarItem.jsx
```diff
@@ -1,59 +1,69 @@
...
-        onClick={onClick}
-      >
-        { children || this.renderIconAndLabel() }
-      </div>
-    );
+    let comp;
+
+    if (linkComponent) {
+      comp = (
+        <div
```
yale: why not just `return <div ... />`?

src/components/SideNavBar/SideNavBarItemGroup.jsx
```diff
@@ -0,0 +1,117 @@
...
+        <div className={style.groupLabelWrapper}>
+          {label}
+        </div>
+      </div>
+    );
+  }
+
+  renderSubItems() {
+    const items = [];
+    forEach(this.props.items, (groupedItem) => {
```
yale: `map` would be perfect here

src/components/SideNavBar/SideNavBarTop.jsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div
+      className={style.top}
+      onClick={onClick}
+    >
+      <div className={style.iconAndLabelContainer}>
+        <div className={style.iconWrapper}>
+          <Icon
+            color="#fff"
+            icon={locationPin}
```
yale: seem to be mixing spacestation-specific stuff in plasma - shouldn't ssv2 be passing in `locationPin` to the component?

src/components/layout/FixedLeft/FixedLeft.jsx
```diff
@@ -1,14 +1,25 @@
...
+import getDataAttrs from '../../../getDataAttrs';
 
 /**
   * The fixed left page is great.
  */
 
+/* eslint-disable */
+const getOffsetTop = (elem) => {
+  let offsetTop = 0;
+  do {
```
yale: whoa. never seen a `do while` in js.

perhaps more idiomatic to use `while (...) {...}`

src/components/layout/FixedRight/FixedRight.jsx
yale: maybe DRY this up, since this is used in two files

src/components/layout/Page/Page.jsx
```diff
@@ -0,0 +1,61 @@
...
+      title,
+      children,
+      actions,
+      breadcrumb,
+    } = this.props;
+
+    return (
+      <div
+        className={style.wrapper}
+        ref={(c) => { this.wrapper = c; }}
```
yale: when is this needed?

package.json
kangax: Oh shit, you're right. I think I tried upgrading this before and there are a lot of changes in component structure that require us to change most of our custom styles.

src/components/layout/FixedLeft/FixedLeft.jsx
kangax: This looks like a "standard" upwards tree iteration to collect values. Nick probably took it from the web. I've seen before and looks good. I'd just want to move it to a shared method.

src/components/Modal/Modal.jsx
nickstamas: As a style thing, I usually prefer to limit the use of early returns because it makes the code easier to follow (in this very simple case, not a big deal obvi). If there's a strong preference for that style, we can change.

src/components/Card/style.scss
```diff
@@ -27,9 +27,26 @@
 }
 
 .condensed {
-  padding: $space20 $space20 0 $space20;
+  padding: $space30;
```
nickstamas: Yeah, it is.

src/components/Modal/Modal.jsx
nickstamas: Ha, good call. 🤔

src/components/layout/FixedLeft/FixedLeft.jsx
nickstamas: It is syntax that tells the auto-generating doc parser that this it shouldn't render a preview for this component.

src/components/SideNavBar/OldSideNavBar.jsx
nickstamas: Probably but this component will be removed once we make the switch over in Spacestation to the new sidenav.

PR #105, opened 2017-07-06
add size small button prop

aaroncouture:


src/components/Button/Button.jsx
```diff
@@ -7,6 +7,7 @@ import Loader from '../Loader/Loader.jsx';
 import style from './style.scss';
 
 const type = { PRIMARY: 'primary', SECONDARY: 'secondary', TERTIARY: 'tertiary' };
+const SIZE = { SMALL: 'small' };
```
kangax: Why are these in different cases?

PR #113, opened 2017-07-24
Vertical tabs

yale:
screen shot 2017-07-24 at 4 58 56 pm

### DIFF
--- a/src/components/Tabs/style.scss
+++ b/src/components/Tabs/style.scss
@@ -6,6 +6,11 @@
   list-style: none;
 }
 
+.verticalTabs {
+  padding-left: 0px;
+  list-style: none;
+}
+
 .tab {
   display: inline-block;
   @extend %fontFamily--body;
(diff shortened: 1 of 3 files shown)


src/components/Tabs/style.scss
```diff
@@ -43,3 +48,22 @@
     border-bottom: 1px solid $cGray600;
   }
 }
+
+.verticalTab {
```
kangax: maybe vertical styles should just be an addition to "base" tab ones? seems strange that vertical would look inherently different than horizontal as far as things like background color of active tab, font size, etc.

yale: @nickstamas thoughts?

nickstamas:
I don't know, this is a tricky one because as like a human with no other
context, I wouldn't ever describe this as "vertical" tabs, tabs to me are
always horizontal.

But functionally they're incredibly similar. I'd say go with what we talked
about @yale for now and maybe we'll rethink later. It's a good point Juriy.

PR #118, opened 2017-08-28
update table to support sort handler prop

pseuyi:


src/components/Table/Table.jsx
```diff
@@ -166,7 +166,7 @@ class Table extends React.Component {
               this.headerComponents[key] = c;
             }
           }}
-          onClick={value.sortable && value.onClick}
+          onClick={() => value.sortable && this.props.onSort(key)}
```
aaroncouture: add a propType check for onSort

pseuyi: @aaroncouture updated

wingyli: ooo actually if `value.sortable` is true, we should also check to make sure `onSort` is defined before calling the function

PR #126, opened 2017-09-27
fixing actions spacing on modal

yale:
## Before
screen shot 2017-09-27 at 1 59 43 pm

## After
screen shot 2017-09-27 at 1 59 55 pm

### DIFF
--- a/src/components/Modal/style.scss
+++ b/src/components/Modal/style.scss
@@ -70,3 +70,13 @@
 .content {
   @extend %layout--bounds;
 }
+
+.pageActions {
+  display: flex;
+  *  {
+    margin-right: 20px;
+  }
+  :last-child {
+    margin-right: 0;
+  }
+}
--- a/src/components/Modal/style.scss
+++ b/src/components/Modal/style.scss
@@ -73,9 +73,11 @@
 
 .pageActions {
   display: flex;
-  *  {
+
+  * {
     margin-right: 20px;
   }
+
   :last-child {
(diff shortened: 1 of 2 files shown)


src/components/Modal/style.scss
```diff
@@ -70,3 +70,13 @@
 .content {
   @extend %layout--bounds;
 }
+
+.pageActions {
```
yale: this style was being used in the `Modal` component but wasn't defined. I pulled this from the `Page` component.

src/components/Modal/style.scss
```diff
@@ -70,3 +70,13 @@
...
+  display: flex;
+  *  {
```
wingyli: minor: extra spacing here `* {`

rfarine: FIX THE SPACING YALE

src/components/Modal/style.scss
```diff
@@ -70,3 +70,15 @@
 .content {
   @extend %layout--bounds;
 }
+
+.pageActions {
+  display: flex;
+
+  * {
```
kangax: You sure this is not too broad? Performance and future additions (of inside content) might suffer.

rfarine: I was wondering about this too. I'm wondering if we'll ever have other content in the actions.

yale: @nickstamas ^

PR #127, opened 2017-10-12
add new style for Tabs component

pseuyi:
screen shot 2017-10-12 at 5 44 15 pm

### DIFF
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -33,7 +33,7 @@ class Tabs extends React.Component {
     return style.tab;
   }
 
-  getTabClassNames= (tab, index) => {
+  getTabClassNames = (tab, index) => {
     const baseTabClass = this.getTabStyle();
     return cn(baseTabClass, {
       [style.active]: (
@@ -60,11 +60,10 @@ class Tabs extends React.Component {
   }
 
   render() {
-    const tabsStyle = this.getTabsStyle();
     return (
       <ul
         {...getDataAttrs(this.props.data)}
-        className={tabsStyle}
+        className={this.getTabsStyle()}
         role="tablist"
       >
         { map(this.props.items, this.renderTab) }
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -19,9 +19,22 @@ class Tabs extends React.Component {
     }
   }
 
-  getTabClasses = (tab, index) => {
-    const baseTabClass = this.props.vertical ? style.verticalTab : style.tab;
+  getTabsStyle() {
+    const { vertical, plain } = this.props;
+    if (vertical) return style.verticalTabs;
+    if (plain) return style.plainTabs;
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.jsx
```diff
@@ -19,9 +19,22 @@ class Tabs extends React.Component {
...
+  }
+
+  getTabStyle() {
+    const { vertical, plain } = this.props;
+    if (vertical) return style.verticalTab;
+    if (plain) return style.plainTab;
+    return style.tab;
+  }
 
+  getTabClassNames= (tab, index) => {
```
kangax: nit: missing space

src/components/Tabs/Tabs.jsx
```diff
@@ -47,10 +60,11 @@ class Tabs extends React.Component {
   }
 
   render() {
+    const tabsStyle = this.getTabsStyle();
```
kangax: I think no need for variable here since it doesn't add any clarity, can just use inline

src/components/Tabs/style.scss
```diff
@@ -2,7 +2,7 @@
 
 .tabs {
   padding: 14px 0 0 14px; /* so first tab aligns with labels */
-  border-bottom: 1px solid $cGray200;;
+  border-bottom: 1px solid $cGray200;
```
nickstamas: Can we use the new color convention here? I should mark more clearly the deprecated colors, but we are moving to the systematized colors. This would probably map to $cGray40.

src/components/Tabs/style.scss
```diff
@@ -67,3 +67,28 @@
     background-color: $cWhite300;
   }
 }
+
+.plainTabs {
+  padding: 14px 0;
+  border-bottom: 1px solid $cGray200;
```
nickstamas: $cGray40

src/components/Tabs/style.scss
```diff
@@ -67,3 +67,28 @@
...
+  list-style: none;
+}
+
+.plainTab {
+  display: inline-block;
+  @extend %fontFamily--body;
+  font-size: 14px;
+  color: $cWeWorkBlack;
```
nickstamas: $cBlack50

src/components/Tabs/style.scss
pseuyi: def, good to know

src/components/Tabs/Tabs.jsx
```diff
@@ -19,9 +19,22 @@ class Tabs extends React.Component {
...
-  getTabClasses = (tab, index) => {
-    const baseTabClass = this.props.vertical ? style.verticalTab : style.tab;
+  getTabsStyle() {
+    const { vertical, plain } = this.props;
+    if (vertical) return style.verticalTabs;
+    if (plain) return style.plainTabs;
+    return style.tabs;
+  }
+
+  getTabStyle() {
```
kangax: oh wow, for a second I thought there were duplicate methods

PR #131, opened 2017-11-14
DMX-4218 Plasma Checkbox fixes

csakai:
We want to use plasma checkbox and allow for disabled look+feel
* Add classnames to Checkbox.jsx
* Add disabled styling for Checkbox.jsx
* Apply inline-flex to label container (and separate from description)
* The description was stretching out the containing div, causing the label's clickable area to be much larger than the text that should be clickable
Also, remove extra declaration for type on FormField's Label

UI changes:
Old Original Checkbox (label clickable-area is way past the text):  
[Original Before click](https://user-images.githubusercontent.com/3250676/32805185-bc494cae-c956-11e7-9767-72a5f1e7fdce.jpeg)
[Original After click](https://user-images.githubusercontent.com/3250676/32805190-bcff0ada-c956-11e7-8b1e-1da2089c00a7.jpeg)
Old Indeterminate Checkbox (label clickable-area is way past the text):  
[Indeterminate Before click](https://user-images.githubusercontent.com/3250676/32805191-bd087bd8-c956-11e7-91ab-a47af9ce0f22.jpeg)
[Indeterminaste After click](https://user-images.githubusercontent.com/3250676/32805192-bd135422-c956-11e7-898a-e96ca7e05f51.jpeg)
New Original Checkbox (note, label clickable area does not extend past the text):  
[plasma dmx-4218 notclickable](https://user-images.githubusercontent.com/3250676/32805193-bd22b43a-c956-11e7-866a-45977f09349a.jpeg)
New Indeterminate Checkbox (note, label clickable area does not extend past the text):  
[plasma dmx-4218 notclickableindeterminate](https://user-images.githubusercontent.com/3250676/32805194-bd3159e0-c956-11e7-985b-4659ca100093.jpeg)

I demonstrate the changes to the indeterminate checkbox as well because I needed to add spacing to the "unchecked" state of the fake checkbox due to the introduction of new elements shifting the position of the label upon check/uncheck.

Also, note that I am able to set the label text to be bold via props - The designs I was given call for a bold label, but I can see why you would want checkboxes to not necessarily have a bold label.

I've also linked two mov files that demonstrate the disabled styling changes:  
[Old Disabled Checkbox Behavior](https://drive.google.com/open?id=1G0Wpx4eNA24URpFqbSdcxLSbRsDDHYyA)
[New Disabled Checkbox Behavior](https://drive.google.com/open?id=1J31gWcaAWQHNtalUutK_uNlE0a2ScyTC)

### DIFF
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -24,13 +24,11 @@ class Checkbox extends React.Component {
     } = this.props;
     const id = uniqueId('id');
     const inputClassName = indeterminate ? style.indeterminate : style.original;
-    const wrapperClassName = cx({
-      [style.wrapper]: !disabled,
+    const wrapperClassName = cx(style.wrapper, {
       [style.wrapperDisabled]: disabled,
     });
-    const textClassName = cx({
+    const textClassName = cx(style.text, {
       [style.textBold]: isBold,
-      [style.text]: !disabled,
       [style.textDisabled]: disabled,
     });
 
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -1,6 +1,7 @@
 import { uniqueId } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 import style from './style.scss';
 import {
   getDataAttrs,
@@ -11,6 +12,7 @@ class Checkbox extends React.Component {
 
   render() {
     const {
+      isBold,
       checked,
       indeterminate,
       name,
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
     } = this.props;
     const id = uniqueId('id');
     const inputClassName = indeterminate ? style.indeterminate : style.original;
+    const wrapperClassName = cx({
+      [style.wrapper]: !disabled,
+      [style.wrapperDisabled]: disabled
+    });
+    const textClassName = cx({
+      [style.textBold]: isBold,
+      [style.text]: !disabled,
+      [style.textDisabled]: disabled
+    });
 
     return (
       <div
         {...getDataAttrs(data)}
       >
-        <label
-          htmlFor={id}
-          className={style.wrapper}
-        >
-          <input
-            disabled={disabled}
-            checked={checked || value}
-            className={inputClassName}
-            type="checkbox"
-            id={id}
-            name={name}
-            onChange={onChange}
-          />
-          <div className={style.checkbox} />
-          <div className={style.text}>
-            {text}
-          </div>
-        </label>
+        <div>
+          <label
+            htmlFor={id}
+            className={wrapperClassName}
+          >
+            <input
+              disabled={disabled}
+              checked={checked || value}
+              className={inputClassName}
+              type="checkbox"
+              id={id}
+              name={name}
+              onChange={onChange}
+            />
+            <div className={style.checkbox} />
+            <div className={textClassName}>
+              {text}
+            </div>
+          </label>
+        </div>
         { this.props.description &&
           <span className={style.description}>
             { this.props.description }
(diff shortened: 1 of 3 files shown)


src/components/Checkbox/Checkbox.jsx
```diff
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
...
-            {text}
-          </div>
-        </label>
+        <div>
+          <label
+            htmlFor={id}
+            className={wrapperClassName}
+          >
+            <input
+              disabled={disabled}
```
nickstamas: This should use the shorthand: 'disabled' (i.e., without '={disabled}')

src/components/Checkbox/Checkbox.jsx
```diff
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
...
+              checked={checked || value}
+              className={inputClassName}
+              type="checkbox"
+              id={id}
```
nickstamas: Use shorthand

src/components/Checkbox/Checkbox.jsx
```diff
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
...
+              name={name}
```
nickstamas: Use shorthand

src/components/Checkbox/Checkbox.jsx
```diff
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
...
+              onChange={onChange}
```
nickstamas: Use shorthand

src/components/Checkbox/Checkbox.jsx
```diff
@@ -22,29 +24,40 @@ class Checkbox extends React.Component {
...
     const inputClassName = indeterminate ? style.indeterminate : style.original;
+    const wrapperClassName = cx({
+      [style.wrapper]: !disabled,
+      [style.wrapperDisabled]: disabled,
+    });
+    const textClassName = cx({
+      [style.textBold]: isBold,
+      [style.text]: !disabled,
+      [style.textDisabled]: disabled,
+    });
```
rfarine: Why don't you just do something like

```
cx(style.text, {
  [style.textBold]: isBold,
  [style.textDisabled]: disabled,
})
```

where `.text` would just be:

```
margin-left: 10px;
@extend %type--body;		   
cursor: pointer;
```

and you'd get rid if `.textBase` cause that's unnecessary

and then `.textDisabled` would become

```
color: $cGray200;
```

and if need be, add the disabled cursor to that last one?

src/components/Checkbox/style.scss
```diff
@@ -1,27 +1,46 @@
...
   }
 }
 
+.wrapperDisabled {
+  composes: wrapperBase;
+  cursor: not-allowed;
+  .original, .indeterminate {
+    cursor: not-allowed;
+  }
+}
```
rfarine: See comments above -- same sort of thing could be done with wrapper here.

src/components/Checkbox/Checkbox.jsx
csakai: sounds good

csakai: not sure what my motivation was, separating them into separate classes that each had the same thing, maybe it was a BEM thing 🤷‍♂️

csakai: oh! now I remember, I didn't want to have the hover styling for the checkbox applying when the wrapper was set to disabled, and I didn't want to have to override that style with the default style. But I can definitely remove the repetition on the text classes

src/components/Checkbox/style.scss
```diff
@@ -19,19 +19,22 @@
...
 
 .wrapperDisabled {
-  composes: wrapperBase;
   cursor: not-allowed;
   .original, .indeterminate {
     cursor: not-allowed;
   }
+  &:hover {
+    .checkbox {
+      border: $checkboxBorder;
```
rfarine: Just curious - why did you turn this into a variable?

PR #148, opened 2018-01-24
Version 1.0 with React 16 support

kangax:


kangax:
@aaroncouture Yale brought a good point of just referring to this branch in SS package.json; that way we can keep all the plasma master changes in both places until we make a final switch.

PR #151, opened 2018-03-07
CMOS 3607 fancy radio buttons

pseuyi:
[cmos 3607](https://jira.we.co/browse/CMOS-3607)

cmos 3607 fancy radio buttons


src/components/RadioButton/RadioButton.jsx
```diff
@@ -12,60 +13,79 @@ class RadioButton extends React.Component {
   };
 
   render() {
```
pseuyi: ordered les props

PR #152, opened 2018-03-07
Add maxLength to input and textarea

kangax:


stories/TextArea.js
```diff
@@ -20,4 +20,9 @@ storiesOf('TextArea', module)
     <div>
       <TextArea disabled placeholder="Type a note..." />
     </div>
+  ))
+  .add('maxLength', () => (
+    <div>
+      <TextArea maxLength="10" placeholder="Max length of 10" />
```
pseuyi: so this won't show the max anywhere? it will just be in the input?

kangax: yeah, it basically just passes maxLength to the underlying native element — https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-maxlength

kangax: I'm adding another PR to add a character counter information that will respect maxLength as well

PR #154, opened 2018-03-08
[RFC] Add withCounter option to textarea

kangax:
This makes textarea a controlled component and gives it a `withCounter` option (with ability to customize it via `counterStyle`)

### withCounter

screenshot 2018-03-08 15 13 12

### withCounter + maxLength

screenshot 2018-03-08 15 13 49

### DIFF
--- a/StoryState.js
+++ b/StoryState.js
@@ -1,0 +1,14 @@
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+
+export default class StoryState extends Component {
+  static propTypes = {
+    render: PropTypes.func.isRequired,
+  };
+
+  state = {};
+
+  render() {
+    return this.props.render(this.state, this.setState.bind(this));
+  }
+}
--- a/src/components/TextArea/TextArea.jsx
+++ b/src/components/TextArea/TextArea.jsx
@@ -7,31 +7,19 @@ import Autogrow from './autogrow';
 import style from './style.scss';
 
 class TextArea extends Component {
-  onChange = (...args) => {
-    this.setState({
-      value: this.textarea.value,
-    });
-    return this.props.onChange && this.props.onChange(...args);
-  };
-  constructor(props) {
-    super(props);
-    this.state = {
-      value: this.props.value || '',
-    };
-  }
   render() {
     const {
-      counterStyle,
       data,
       disabled,
       error,
       maxLength,
       onBlur,
+      onChange,
       onFocus,
       placeholder,
       rows,
       size,
-      withCounter,
+      value,
     } = this.props;
 
     const wrapperStyle = cx(style.wrapper, {
@@ -47,18 +35,17 @@ class TextArea extends Component {
 
     return (
       <div>
-        {withCounter && (
-          <div className={counterStyle || style.counterStyle}>
-            {this.state.value.length}
-            {maxLength && `/${maxLength}`} characters
-          </div>
-        )}
         <div className={wrapperStyle} {...getDataAttrs(data)}>
           <textarea
-            ref={el => el && (this.textarea = el) && new Autogrow(el)}
+            ref={el => {
+              if (el && !el.props) {
+                new Autogrow(el);
+                this.props.parentRef && this.props.parentRef(el);
+              }
+            }}
             className={textareaStyle}
             disabled={disabled}
-            onChange={this.onChange}
+            onChange={onChange}
             rows={rows}
             onFocus={e => {
               /* eslint-disable no-param-reassign */
@@ -71,7 +58,7 @@ class TextArea extends Component {
               onBlur && onBlur();
             }}
             placeholder={placeholder}
-            value={this.state.value}
+            value={value}
             maxLength={maxLength}
           />
         </div>
@@ -81,7 +68,6 @@ class TextArea extends Component {
 }
 
 TextArea.propTypes = {
-  counterStyle: PropTypes.object,
   disabled: PropTypes.bool,
   error: PropTypes.bool,
   maxLength: PropTypes.string,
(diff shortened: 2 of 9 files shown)


src/components/TextArea/TextArea.jsx
```diff
@@ -8,55 +8,74 @@ import style from './style.scss';
...
+      [style.large]: toUpper(size) === 'LARGE',
     });
 
     return (
-      <div className={wrapperStyle} {...getDataAttrs(this.props.data)}>
+      <div className={wrapperStyle} {...getDataAttrs(data)}>
         <textarea
-          ref={el => el && new Autogrow(el)}
+          ref={el => {
+            if (el && !el.props) {
```
aaroncouture: why would el have props? its a dom element?

kangax: it's either dom element (which we need) or an instance; i'm not sure why React does this but it seems to interchange in my console

src/decorators/withCounter/withCounter.jsx
```diff
@@ -0,0 +1,62 @@
...
+      super(props);
+
+      this.state = {
+        value: props.value || '',
+      };
+    }
+
+    onChange = () => {
+      this.setState({
+        value: this.element.value || '',
```
aaroncouture: why do we need to do this?

src/decorators/withCounter/withCounter.jsx
```diff
@@ -0,0 +1,62 @@
...
+      ref: PropTypes.function,
+      counterStyle: PropTypes.object,
+      maxLength: PropTypes.string,
+    };
+
+    constructor(props) {
+      super(props);
+
+      this.state = {
+        value: props.value || '',
```
aaroncouture: why do we need state at all?

src/decorators/withCounter/withCounter.jsx
```diff
@@ -0,0 +1,62 @@
...
+          };
+        })(props.onChange);
+      } else {
+        props.onChange = this.onChange;
+      }
+
+      return (
+        <div>
+          <div>
+            {this.state.value.length}
```
aaroncouture: get(this.props.value, 'length')   ???

src/decorators/withCounter/withCounter.jsx
```diff
@@ -0,0 +1,62 @@
...
+        ...this.props,
+        parentRef: el => {
+          if (el) {
+            this.element = el;
+          }
+        },
+        value: this.state.value,
+      };
+
+      if (props.onChange) {
```
aaroncouture: this should be moved out to a class property

src/decorators/withCounter/withCounter.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const withCounter = Component => {
```
yale: it looks like this can just be a regular component, i.e.:
```jsx
<WithCounter value={} counterStyle={} maxLength={}>
  <FormInput ... />
</WithCounter>
```

if you change the `<Component ...` in the `render` function to just `this.props.children`

kangax: what are the pros/cons of HOC vs. standalone component?

aaroncouture: Unfortunately Yale’s example would not work with our current impl of createGotmField... but would work with my render props POC! :D

aaroncouture: **createFormField

aaroncouture: If we merge the POC I’d prefer not using HOCs

kangax: How do we decide when to use HOC and when to use component? E.g. `authenticated(...)` vs. `<Authenticated>...</Authenticated>`

kangax: Or are we just generally moving away from HOC's?

aaroncouture: Another option is to convert withCounter to a render prop

aaroncouture: I think we should be migrating away from HOCs bth

StoryState.js
```diff
@@ -0,0 +1,14 @@
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+
+export default class StoryState extends Component {
```
pseuyi: interesting..

src/decorators/withCounter/withCounter.jsx
pseuyi: ^ agree
i'm hoping we can migrate away from the `authenticated()` hoc towards `<Route render={...}/>`

src/components/TextArea/TextArea.jsx
```diff
@@ -8,55 +8,69 @@ import style from './style.scss';
 
 class TextArea extends Component {
   render() {
+    const {
```
aaroncouture: did anything actually change in this file besides the spacing/destructuring?

stories/TextArea.jsx
```diff
@@ -0,0 +1,57 @@
...
+    </div>
+  ))
+  .add('maxLength', () => (
+    <div>
+      <TextArea maxLength="10" placeholder="Max length of 10" />
+    </div>
+  ))
+  .add('withCounter', () => (
+    <div>
+      <StoryState
```
aaroncouture: just realized we can call this WithState :p

aaroncouture (approved):

wingyli (approved):

PR #155, opened 2018-03-13
update number input

pseuyi:


src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,57 +1,69 @@
...
-    this.onPressDecrement = this.onPressDecrement.bind(this);
-    this.onChange = this.onChange.bind(this);
-  }
-
-  onPressIncrement() {
-    if (!isNumber(this.props.maxValue) ||
-        this.props.maxValue >= this.props.value + this.props.step) {
-      this.props.onPressIncrement({ diff: this.props.step });
+  onPressIncrement = () => {
+    const { maxValue, value, step, onPressIncrement, onBlur, onChange } = this.props;
```
aaroncouture: onPressIncrement is not referenced

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,57 +1,69 @@
...
+      onBlur && onBlur();
     }
   }
 
-  onPressDecrement() {
-    if (!isNumber(this.props.minValue) ||
-        this.props.minValue <= this.props.value - this.props.step) {
-      this.props.onPressDecrement({ diff: -1 * this.props.step });
+  onPressDecrement = () => {
+    const { minValue, value, step, onPressDecrement, onBlur, onChange } = this.props;
```
aaroncouture: onPressDecrement is not referenced

pseuyi: yaya

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,57 +1,69 @@
...
-      (!this.props.minValue || value >= this.props.minValue)
-    ) {
-      this.props.onChange({ value });
-    }
+  onChange = (event) => {
+    const { maxValue, minValue, onChange } = this.props;
+    const eventValue = event.nativeEvent.target.value;
+
+    // if the value is cleared, do not convert empty string to 0
+    const value = eventValue === "" ? null : toNumber(eventValue);
```
aaroncouture: dont you just want to return null if empty string?

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,57 +1,69 @@
...
-          placeholder={this.props.placeholder}
-          className={style.input}
-          value={this.props.value}
+          placeholder={placeholder}
+          className={inputStyle}
+          disabled={disabled}
+          value={value}
           onChange={this.onChange}
+          onBlur={(e) => {
+            onBlur && onBlur();
```
aaroncouture: we should not pass anonymous functions if possible.
i think you can just pass onBlur={this.props.onBlur}

src/components/NumberInput/NumberInput.jsx
```diff
@@ -69,20 +81,24 @@ class NumberInput extends React.Component {
...
+  onChange: PropTypes.func.isRequired,
+  onFocus: PropTypes.func,
+  placeholder: PropTypes.string,
   step: PropTypes.number,
-  value: PropTypes.number,
-  onChange: PropTypes.func,
-  onPressIncrement: PropTypes.func,
-  onPressDecrement: PropTypes.func,
-  ...getDataProps(),
+  value: PropTypes.oneOfType([ PropTypes.string, PropTypes.number ]).isRequired,
```
aaroncouture: why can value be a string?

aaroncouture: im assuming this is because some places are passing string into the input? (ie. non-redux-form usages)

src/components/NumberInput/NumberInput.jsx
aaroncouture: nm this is ok

aaroncouture: this means you cant null out a field if there is a minimum value

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,57 +1,69 @@
...
-  onChange(event) {
-    const value = toNumber(event.nativeEvent.target.value);
-    if (
-      (!this.props.maxValue || value <= this.props.maxValue) &&
-      (!this.props.minValue || value >= this.props.minValue)
-    ) {
-      this.props.onChange({ value });
-    }
+  onChange = (event) => {
+    const { maxValue, minValue, onChange } = this.props;
```
aaroncouture: another thought, if minValue is > 9 I dont think you can enter any text :)

src/components/NumberInput/NumberInput.jsx
pseuyi: that is what is happening here. `onChange` is called with `null`, setting the field value to `null` instead of `0`

src/components/NumberInput/NumberInput.jsx
pseuyi: ah yea..this will be needed if we want the custom styles but can remove for now

src/components/NumberInput/NumberInput.jsx
aaroncouture: seems like you need to add a check somewhere to see if minimumValue is defined assuming we ant to be able null out the field.

aaroncouture:
Awesome work :) 👍

src/components/NumberInput/NumberInput.jsx
aaroncouture: we should probably add a prop called allowBlank that if the value is null, we skip these checks.

wingyli (approved):

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,88 +1,125 @@
...
-    ) {
-      this.props.onChange({ value });
-    }
+  handleBlur = e => {
+    e.target.parentElement.classList.remove(style.wrapperFocused);
+    this.props.onBlur && this.props.onBlur();
+  }
+
+  handleFocus = e => {
+    e.target.parentElement.classList.add(style.wrapperFocused);
```
aaroncouture: what do we need this for?

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,88 +1,125 @@
...
   }
 
-  onPressDecrement() {
-    if (!isNumber(this.props.minValue) ||
-        this.props.minValue <= this.props.value - this.props.step) {
-      this.props.onPressDecrement({ diff: -1 * this.props.step });
-    }
+  handleChange = e => {
+    const { onChange, maxValue, minValue } = this.props;
+    //when input is cleared, the default value will be 0
```
aaroncouture: add a check here to see if clearable, and return null if the field is blank.

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,88 +1,125 @@
...
+  onChange: PropTypes.func.isRequired,
+  onFocus: PropTypes.func,
+  placeholder: PropTypes.string,
   step: PropTypes.number,
-  value: PropTypes.number,
-  onChange: PropTypes.func,
-  onPressIncrement: PropTypes.func,
-  onPressDecrement: PropTypes.func,
-  ...getDataProps(),
+  value: PropTypes.oneOfType([emptyStringType, PropTypes.number]).isRequired,
```
aaroncouture: we can get rid of the emptyStringType

src/components/NumberInput/NumberInput.jsx
aaroncouture: after some thought, Im fine with this.

src/components/NumberInput/NumberInput.jsx
```diff
@@ -1,88 +1,125 @@
...
+    const wrapperStyle = cx(style.wrapper, {
+      [style.wrapperDisabled]: disabled,
+      [style.wrapperError]: error,
+    })
+
+    const inputStyle = cx(style.input, {
+      [style.inputDisabled]: disabled,
+    })
+
+    const displayValue = clearable && value === 0 ? "" : value;
```
aaroncouture: dont forget to change this too ;)

aaroncouture (approved):

PR #158, opened 2018-03-27
update node version to 8.9.3

aaroncouture:


kangax:
Technically it would be consistent with SS:

screenshot 2018-03-27 17 36 49

PR #161, opened 2018-03-28
Added red border color option to Card component

chelseawework:
This change relates to ban more information feature: https://jira.we.co/browse/CMOS-3542

Card Red Border:
storybook_card_redborder

Collapsible Card:
screen shot 2018-03-29 at 11 13 44 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.idea
 
 node_modules
 dist
--- a/src/components/Card/Card.jsx
+++ b/src/components/Card/Card.jsx
@@ -44,6 +44,7 @@ class Card extends React.Component {
       data,
       isExpanded,
       borderType,
+      borderColorRed,
       type,
       minWidth,
     } = this.props;
@@ -51,6 +52,7 @@ class Card extends React.Component {
     const cardStyle = cx(style.card, {
       [style.condensed]: type === cardTypes.CONDENSED,
       [style.borderDashed]: borderType === borderTypes.DASHED,
+      [style.borderColorRed]: borderColorRed,
       [style.isExpanded]: isExpanded,
     });
 
@@ -119,6 +121,7 @@ Card.propTypes = {
   type: PropTypes.oneOf([cardTypes.REGULAR, cardTypes.CONDENSED]),
   onClick: PropTypes.func,
   borderType: PropTypes.string,
+  borderColorRed: PropTypes.bool,
   expandedComponent: PropTypes.node,
   isExpanded: PropTypes.bool,
   minWidth: PropTypes.number,
--- a/src/components/Card/Card.jsx
+++ b/src/components/Card/Card.jsx
@@ -12,6 +12,7 @@ const cardTypes = {
   CONDENSED: 'condensed',
 };
 const borderTypes = { DASHED: 'dashed' };
+const borderColors = { RED: 'red' };
 
 class Card extends React.Component {
   constructor() {
@@ -44,7 +45,7 @@ class Card extends React.Component {
       data,
       isExpanded,
       borderType,
-      borderColorRed,
+      borderColor,
       type,
       minWidth,
     } = this.props;
@@ -52,7 +53,7 @@ class Card extends React.Component {
     const cardStyle = cx(style.card, {
       [style.condensed]: type === cardTypes.CONDENSED,
       [style.borderDashed]: borderType === borderTypes.DASHED,
-      [style.borderColorRed]: borderColorRed,
+      [style.borderColorRed]: borderColor === borderColors.RED,
       [style.isExpanded]: isExpanded,
     });
 
--- a/src/components/Card/Card.jsx
+++ b/src/components/Card/Card.jsx
@@ -12,7 +12,7 @@ const cardTypes = {
   CONDENSED: 'condensed',
 };
 const borderTypes = { DASHED: 'dashed' };
-const borderColors = { RED: 'red' };
+export const borderColors = { RED: 'red' };
 
 class Card extends React.Component {
   constructor() {
@@ -53,7 +53,7 @@ class Card extends React.Component {
     const cardStyle = cx(style.card, {
       [style.condensed]: type === cardTypes.CONDENSED,
       [style.borderDashed]: borderType === borderTypes.DASHED,
-      [style.borderColorRed]: borderColor === borderColors.RED,
+      [style.borderRed]: borderColor === borderColors.RED,
       [style.isExpanded]: isExpanded,
     });
 
@@ -122,7 +122,7 @@ Card.propTypes = {
   type: PropTypes.oneOf([cardTypes.REGULAR, cardTypes.CONDENSED]),
   onClick: PropTypes.func,
   borderType: PropTypes.string,
-  borderColorRed: PropTypes.bool,
+  borderColor: PropTypes.oneOf([borderColors.RED]),
   expandedComponent: PropTypes.node,
   isExpanded: PropTypes.bool,
   minWidth: PropTypes.number,
--- a/src/components/CollapsibleCard/CollapsibleCard.jsx
+++ b/src/components/CollapsibleCard/CollapsibleCard.jsx
@@ -14,11 +14,16 @@ class CollapsibleCard extends Component {
     expandableText: PropTypes.node,
     children: PropTypes.node.isRequired,
     styleColor: PropTypes.oneOf([styleColors.RED]),
+    isCollapsed: PropTypes.bool,
     disabled: PropTypes.bool,
   };
 
+  static defaultProps = {
+    isCollapsed: false,
+  }
+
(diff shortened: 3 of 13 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
+.idea
```
kangax: is this something your IDE uses?

src/components/Card/Card.jsx
```diff
@@ -44,13 +44,15 @@ class Card extends React.Component {
       data,
       isExpanded,
       borderType,
+      borderColorRed,
```
kangax: this should be borderColor, similar to borderType, so that we don't need to add other props later like borderColorGreen, borderColorBlue and so on. Ideally, it'd just be something like borderStyle (for even more flexibility) but that would be backward-incompatible so we can revisit it later.

.gitignore
chelseawework: Yes, JetBrains creates a `.idea` directory. I didn't have issues with spacestation because this line is already in the `.gitignore`.

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+            </Card>
+        )}
+      </div>
+    );
+  }
+}
+
+CollapsibleCard.defaultProps = {};
+
+CollapsibleCard.propTypes = {
```
kangax: do we have static props support in plasma? would be better to use those

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+import Card from '../Card/Card';
+
+// Style
+import style from './style.scss';
+
+const styleColors = { RED: 'red' };
+
+class CollapsibleCard extends Component {
+  state = {
+    displayCard: false,
```
kangax: let's use `shouldDisplayCard` since that's a convention for booleans (`displayCard` is a verb and suggests that it's a method that displays a card)

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+      }
+    );
+
+    return (
+      <div className={collapsibleCardStyle}>
+        <div className={style.cardHeader}>
+          {headerText}
+          {!disabled && (
+            <div className={style.textLink} onClick={this.onClick}>
+              <b>{expandableText}</b>
```
kangax: Should the text always be bold?

src/components/CollapsibleCard/CollapsibleCard.jsx
kangax: do we want collapsibility to be controlled by props too? e.g. if the component is connected (in that case we'd need to support passing it via props)

src/components/Card/Card.jsx
```diff
@@ -119,6 +122,7 @@ Card.propTypes = {
   type: PropTypes.oneOf([cardTypes.REGULAR, cardTypes.CONDENSED]),
   onClick: PropTypes.func,
   borderType: PropTypes.string,
+  borderColorRed: PropTypes.bool,
```
aaroncouture: shouldnt this be borderColor: PropTypes.string

src/components/Card/Card.jsx
```diff
@@ -44,13 +45,15 @@ class Card extends React.Component {
...
       borderType,
+      borderColor,
       type,
       minWidth,
     } = this.props;
 
     const cardStyle = cx(style.card, {
       [style.condensed]: type === cardTypes.CONDENSED,
       [style.borderDashed]: borderType === borderTypes.DASHED,
+      [style.borderColorRed]: borderColor === borderColors.RED,
```
wingyli: minor: rename the style to `borderRed`

src/components/Card/Card.jsx
wingyli: prop name should be `borderRed` and the prop type can be changed to `PropTypes.oneOf([borderColors.RED])` instead of just `string`

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+// Components
+import Card from '../Card/Card';
+
+// Style
+import style from './style.scss';
+
+const styleColors = { RED: 'red' };
```
wingyli: maybe this can be imported from card?

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+class CollapsibleCard extends Component {
+  state = {
+    displayCard: false,
+  }
+
+  onClick = () => {
+    this.setState({ displayCard: !this.state.displayCard });
```
wingyli: `this.setState(prevState => ({ displayCard: !prevState. displayCard }));`

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+                className={
+                  ((displayCard) ? style.textLinkArrowUp : style.textLinkArrowDown)
+                }
+              />
+            </div>
+          )}
+        </div>
+        {!disabled &&
+          displayCard && (
+            <Card borderColor={(styleColor === styleColors.RED) && styleColor}>
```
aaroncouture: why not just pass styleColor?

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+    return (
+      <div className={collapsibleCardStyle}>
+        <div className={style.cardHeader}>
+          {headerText}
+          {!disabled && (
+            <div className={style.textLink} onClick={this.onClick}>
+              <b>{expandableText}</b>
+              <div
+                className={
+                  ((displayCard) ? style.textLinkArrowUp : style.textLinkArrowDown)
```
wingyli: these parentheses are not needed and should be on the same line `className={displayCard ? style.textLinkArrowUp : style.textLinkArrowDown}`

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+CollapsibleCard.defaultProps = {};
+
+CollapsibleCard.propTypes = {
+  headerText: PropTypes.node.isRequired,
+  expandableText: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  styleColor: PropTypes.string,
```
wingyli: maybe the prop type here can also be imported from card

src/components/CollapsibleCard/CollapsibleCard.jsx
wingyli: we can just pass thru styleColor prop like `<Card borderColor={styleColor}>` since they should always have the same color

src/components/CollapsibleCard/CollapsibleCard.jsx
chelseawework: That is a good point; it may not need to always be bold. I'll remove expandable text always being bold. If it needs to be bold, it can be passed into the component that way.

src/components/Card/Card.jsx
aaroncouture: ahh good point

src/components/CollapsibleCard/CollapsibleCard.jsx
chelseawework: We talked about this earlier. I added a prop to determine if the collapsible element is displayed on first render.

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+// Components
+import Card, { borderColors as styleColors } from '../Card/Card';
+
+// Style
+import style from './style.scss';
+
+class CollapsibleCard extends Component {
+  static propTypes = {
+    headerText: PropTypes.node.isRequired,
+    expandableText: PropTypes.node,
```
wingyli: minor: prefer `collapsibleText`

src/components/CollapsibleCard/CollapsibleCard.jsx
```diff
@@ -0,0 +1,76 @@
...
+      styleColor,
+      disabled,
+    } = this.props;
+
+    const {
+      shouldDisplayCard,
+    } = this.state;
+
+    const collapsibleCardStyle = cx(style.collapsibleCard, {
+      [style.styleColorRed]: styleColor === styleColors.RED,
```
pseuyi: nit - `themeRed` may be clearer

PR #163, opened 2018-04-06
Announcement styles

pseuyi:
- adds `<Text />` to stories with some styles (for use as "hint text")
- adds more flow configuration
- adds `maxWidth` to `<Modal/>`
- adds a default color for the `withCounter` decorator

screen shot 2018-04-06 at 3 37 24 pm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@wework-dev/plasma",
-  "version": "0.0.146",
+  "version": "0.0.147",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/components/Modal/Modal.jsx
+++ b/src/components/Modal/Modal.jsx
@@ -17,6 +17,7 @@ class Modal extends React.Component {
       children,
       onDismiss,
       minWidth = 0,
+      maxWidth,
     } = this.props;
     if (this.props.visible) {
       comp = (
@@ -24,9 +25,9 @@ class Modal extends React.Component {
           {...getDataAttrs(data)}
           className={style.wrapper}
         >
-          <div className={style.innerWrapper}>
+          <div className={style.innerWrapper}yar>
             <div
-              style={{ minWidth }}
+              style={{ minWidth, maxWidth }}
               className={style.card}
             >
               <div className={style.content}>
@@ -67,6 +68,7 @@ Modal.propTypes = {
   onDismiss: PropTypes.func,
   style: PropTypes.object,
   minWidth: PropTypes.number,
+  maxWidth: PropTypes.number,
   ...getDataProps(),
 };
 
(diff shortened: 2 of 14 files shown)


src/components/Modal/Modal.jsx
```diff
@@ -17,16 +17,17 @@ class Modal extends React.Component {
...
+      maxWidth,
     } = this.props;
     if (this.props.visible) {
       comp = (
         <div
           {...getDataAttrs(data)}
           className={style.wrapper}
         >
-          <div className={style.innerWrapper}>
+          <div className={style.innerWrapper}yar>
```
aaroncouture: Yar?

src/components/Text/Text.jsx
```diff
@@ -1,16 +1,39 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import cn from 'classnames';
+// @flow
+/* eslint react/prop-types: 0 */
+import * as React from 'react';
```
aaroncouture: Why are we importing all React?

pseuyi: for flow https://flow.org/en/docs/react/children/

src/components/Modal/Modal.jsx
pseuyi: lol

chelseawework: Should `yar` be here?

src/decorators/withCounter/withCounter.jsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
+import style from './style.scss'
```
chelseawework: Missing simicolon?

src/components/Text/Text.jsx
```diff
@@ -1,16 +1,39 @@
...
+};
+
+const types = {
+  PRIMARY: 'primary',
+  SECONDARY: 'secondary',
+};
+
+class Text extends React.Component<Props> {
+  static defaultProps = {
+    style: { width: 200, height: 'auto' },
```
chelseawework: If you move this to `.scss` file, you could remove `defaultProps` and make this a functional component.

src/decorators/withCounter/withCounter.jsx
pseuyi: ah turned off lint

src/decorators/withCounter/style.scss
```diff
@@ -0,0 +1,5 @@
+@import "../../styles/base";
```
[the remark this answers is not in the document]
pseuyi: will change, unfortunately we have this in many other files...

src/components/Text/Text.jsx
pseuyi: we want this to be in `defaultProps` so that it can be overwritten if `style` is passed in. i don't think we should remove this without refactoring ssv2 code to pass in width rather than assuming a default width.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
```
kangax: I don't think we need package-lock since we already have yarn.lock

kangax: @pseuyi yep, just remove it

PR #166, opened 2018-05-04
Add fancy checkbox styles

chelseawework:
New Fancy Checkboxes:
screen shot 2018-05-04 at 10 13 32 am

New Fancy Checkboxes (with Icon):
screen shot 2018-05-04 at 12 33 36 pm

Jira Ticket: https://jira.we.co/browse/CMOS-3770

### DIFF
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -21,13 +21,14 @@ class Checkbox extends React.Component {
       value,
     } = this.props;
 
+    const isChecked = (checked || value);
     const id = uniqueId('id');
     const inlineClass = cx({ [style.inline]: fancy });
     const inputClassName = indeterminate ? style.indeterminate : style.original;
     const wrapperClassName = cx(style.wrapper, {
       [style.wrapperDisabled]: disabled,
       [style.fancy]: fancy,
-      [style.fancyChecked]: fancy && value,
+      [style.fancyChecked]: fancy && isChecked,
     });
     const textClassName = cx(style.text, {
       [style.textBold]: isBold,
@@ -39,7 +40,7 @@ class Checkbox extends React.Component {
         <div>
           <label htmlFor={id} className={wrapperClassName}>
             <input
-              checked={checked || value}
+              checked={isChecked}
               className={inputClassName}
               disabled={disabled}
               id={id}
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -1,16 +1,42 @@
+// @flow
+/* eslint react/prop-types: 0 */
 import { uniqueId } from 'lodash';
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
 import cx from 'classnames';
 import style from './style.scss';
 import Icon from '../Icon/Icon';
-import { getDataAttrs, getDataProps } from '../../dataUtils';
+import { getDataAttrs } from '../../dataUtils';
+import type { Data } from '../../types';
 
-class Checkbox extends React.Component {
+type Props = {
+  checked?: boolean,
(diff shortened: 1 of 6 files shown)


src/components/Checkbox/Checkbox.jsx
```diff
@@ -8,42 +8,47 @@ import { getDataAttrs, getDataProps } from '../../dataUtils';
 class Checkbox extends React.Component {
   render() {
     const {
-      isBold,
       checked,
+      data,
```
kangax: would be nice if prettier sorted these kind of destructuring for us automatically

kangax:
Let's not forget to update docs after

pseuyi:
did we decide to scrap the little icons on the fancy checkbox design?

src/components/Checkbox/Checkbox.jsx
pseuyi: maybe we should do it all in one go at some pt. ive also been doing it file by file but it can make pr review noisy

pseuyi:
also, we should consider adding an optional style object for a prop since the events mock looks like it will have a custom color scheme...

src/components/Checkbox/style.scss
```diff
@@ -104,6 +132,14 @@ $checkboxBorder: 1px solid $cBorder;
   color: $cGray200;
 }
 
+.icon {
+  margin: 0px -10px 0px 0px;
+}
+
+.fancyIcon {
+  margin: 0px 0px 0px -10px;
```
kangax: nit: 0px --> 0

src/tests/__tests__/Checkbox.react-test.js
```diff
@@ -15,4 +15,14 @@ describe('Checkbox Component', () => {
     tree.props.checked = true;
     expect(tree).toMatchSnapshot();
   });
+
+  test('no fancy style', () => {
+    const component = shallow(<Checkbox>Test Checkbox</Checkbox>);
```
kangax: btw, can't we use snapshots here as well?

chelseawework: Yes, we can. Updated!

src/tests/__tests__/Checkbox.react-test.js
```diff
@@ -15,4 +14,12 @@ describe('Checkbox Component', () => {
     tree.props.checked = true;
     expect(tree).toMatchSnapshot();
   });
+
+  test('render', () => {
+    const component = renderer.create(
+      <Checkbox fancy />
+    );
+    let tree = component.toJSON();
```
kangax: minor: since we're generally avoiding `let` and since the var is not used for anything else AND doesn't really add much to readability, would be best to just `expect(component.toJSON()).toMatchSnapshot()`

src/components/Checkbox/Checkbox.jsx
```diff
@@ -57,22 +70,24 @@ class Checkbox extends React.Component {
...
   checked: false,
-  value: false,
   indeterminate: false,
-  text: 'Option',
   name: 'checkbox',
+  text: 'Option',
+  value: false,
 };
 
 Checkbox.propTypes = {
```
pseuyi: can we add flow? i totally forgot we have that in plasma now

src/components/Checkbox/Checkbox.jsx
```diff
@@ -1,81 +1,100 @@
...
+  description?: string,
+  disabled?: boolean,
+  fancy?: boolean,
+  icon?: string,
+  indeterminate?: boolean,
+  isBold?: boolean,
+  name: string,
+  onBlur?: () => mixed,
+  onChange?: () => mixed,
+  text: string | Node,
```
kangax: `Node` already incudes `string`! :) https://github.com/facebook/flow/blob/master/website/en/docs/react/types.md#toc-react-node

PR #173, opened 2018-06-04
CMOS-3840 Added logic to handle the tab items title being passed as html

chaitany6143:
Added logic in Tabs component to handle the tab items title being passed as html. Please see the attached screenshot.

screen shot 2018-06-04 at 11 02 19 am

### DIFF
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -42,22 +42,33 @@ class Tabs extends React.Component {
       [style.first]: index === 0,
       [style.last]: index === this.props.items.length - 1,
     });
-  }
+  };
 
   renderTab = (tab, index) => {
+    const tabAttrs = {
+      role: 'tab',
+      onClick: () => this.onClick(tab, index),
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.jsx
```diff
@@ -45,6 +45,19 @@ class Tabs extends React.Component {
...
   renderTab = (tab, index) => {
+    if (tab.renderType && tab.renderType === 'html') {
+      return (
+        <li
+          role="tab"
+          onClick={() => this.onClick(tab, index)}
+          onKeyDown={(event) => this.onKeyDown(event, tab, index)}
+          className={this.getTabClassNames(tab, index)}
+          key={index}
+          tabIndex="0"
```
kangax: This looks good, but it'd be nice to get rid of a duplication of props that we pass to `<li>` element. Maybe move them to an object and spread in both places?

chaitany6143: Makes sense! Updated..

pseuyi (approved):

PR #178, opened 2018-07-17
Dismiss modal with esc key

eladitzhakian:


src/components/Modal/Modal.jsx
```diff
@@ -7,7 +7,20 @@ import {
   getDataProps,
 } from '../../dataUtils';
 
+const ESC_KEY_CODE = 27;
+
 class Modal extends React.Component {
+  componentDidMount() {
+    document.addEventListener('keydown', this.dismissOnEscKey, false);
```
kangax: this needs to be cleaned up on unmount

src/components/Modal/Modal.jsx
```diff
@@ -18,8 +31,9 @@ class Modal extends React.Component {
       onDismiss,
       minWidth = 0,
       maxWidth,
+      visible,
     } = this.props;
-    if (this.props.visible) {
+    if (visible) {
```
pseuyi: nice

pseuyi:
don't now if we want to repeat here in plasma but spacestation modals rely on react-hotkeys https://github.com/WeConnect/spacestation-v2/blob/master/package.json#L112

eladitzhakian:
@kangax @pseuyi fixed, please re-review

kangax (commented):
LGTM. The only thing is that, from what I remember, there are cases in SS where modals are specifically not meant to be closed with esc key. Can you please check those and make sure they still work as expected.

eladitzhakian:
@kangax any idea which modals are those?
I could also make this behavior off by default but that doesn't feel right

kangax:
@eladitzhakian it's a `dismissable` prop but I see that it's defined on SS `Modal` which is different from Plasma's `Modal` 🤦‍♂️

PR #182, opened 2018-08-03
[CMOS-4055]: Remove deprecated prop checking from Plasma

chaitany6143:


kangax (commented):
Instead of adding `/* eslint react/prop-types: 0 */` everywhere, we should just disable this rule everwhere in the project

src/components/Button/Button.jsx
```diff
@@ -1,17 +1,34 @@
...
+    children: Node,
+    style: Object,
+    isSubmit: boolean,
+    size: string,
+    data: Data,
+|};
+
+class Button extends React.Component<Props> {
+  static defaultProps = {
+    type: type.PRIMARY,
```
kangax: this actually works?!

src/components/Card/Card.jsx
```diff
@@ -14,7 +16,28 @@ const cardTypes = {
 const borderTypes = { DASHED: 'dashed' };
 export const borderColors = { RED: 'red' };
 
-class Card extends React.Component {
+type Props = {|
+    title: string,
```
kangax: I think our convention is 2 spaces... speaking of which, we need prettier here :/

src/components/Card/Card.jsx
```diff
@@ -25,16 +48,22 @@ class Card extends React.Component {
...
+          = ReactDOM.findDOMNode(this.expanded); //$FlowFixMe
       const expandedHeight = expandedEl.offsetHeight;
       this.setState({ defaultHeight, expandedHeight });
     }
     this.setState({ isMounted: true });
     /* eslint-enable */
   }
 
+  expanded: ?Object;
+  outer: ?Object;
```
kangax: I think these could be marked as Element?

src/components/Header/Header.jsx
```diff
@@ -1,13 +1,28 @@
...
-  getDataAttrs,
-  getDataProps }
-from '../../dataUtils';
+import { getDataAttrs } from '../../dataUtils';
+import type { Data } from '../../types';
 
-class Header extends React.Component {
+type Props = {|
+    text?: string,
+    children: string,
```
kangax: children should always be `Node`

kangax: also could be optional

src/components/Table/Table.jsx
```diff
@@ -1,28 +1,61 @@
...
+
+type State = {|
+    columnSizes: Object,
+    tableWidth: number,
+    backgroundColor?: ?string,
+    stickyCoverHeight?: number,
+    translateY?: number,
+    showSticky?: boolean,
+    headerHeight: number,
+
```
kangax: nit: extra newline

src/components/TextArea/TextArea.jsx
```diff
@@ -1,20 +1,44 @@
...
 
-class TextArea extends Component {
+type Props = {|
+    autosize: boolean,
+    disabled: boolean,
+    error: boolean,
+    maxLength: string,
+    onBlur: () => void,
+    onChange: () => void,
+    onFocus: PropTypes.func,
```
kangax: plz get rid of PropTypes completely

src/components/TextInput/TextInput.jsx
```diff
@@ -1,10 +1,28 @@
...
+    placeholder: string,
+    value: string,
+    onChange: () => void,
+    onFocus: () => void,
+    onBlur: () => void,
+    suffix: Node,
+    prefix: Node,
+    isDisabled: boolean,
+    disabled: boolean,
+    size: boolean,
```
kangax: shouldn't size be a string or a number?

src/components/Toggle/Toggle.jsx
```diff
@@ -14,12 +13,31 @@ const type = {
   SMALL: 'small',
 };
 
-class Toggle extends React.Component {
-  onClick(toggle, index) {
+type Props = {|
+    items: {
+        label: string,
+        title: string,
```
kangax: incorrect spacing here as well

src/components/layout/FixedTop/FixedTop.jsx
```diff
@@ -1,12 +1,27 @@
...
 import style from './style.scss';
 
-class FixedTop extends React.Component {
+
+type State = {|
+    fixedHeight: number,
+|};
+
+type Props = {|
+    children: Array<Node>,
```
kangax: just `Node`

src/components/layout/Group/Group.jsx
```diff
@@ -1,14 +1,25 @@
...
-  getDataAttrs,
-  getDataProps,
-} from '../../../dataUtils';
+import { getDataAttrs } from '../../../dataUtils';
+import type { Data } from '../../../types';
 import style from './style.scss';
 
-class Group extends React.Component {
+type Props = {|
+    children: Array<Node>,
```
kangax: just `Node`

src/components/Button/Button.jsx
chaitany6143: It does. But Changing it to buttonType for clarity.

src/components/Header/Header.jsx
chaitany6143: oops. Changing.

src/components/layout/Group/Group.jsx
chaitany6143: Have to use Array<Node> as this.props.children.length is being accessed. Flow complains on using just Node.

kangax (approved):
🔥

PR #185, opened 2018-08-03
[CMOS-4057] Add flow to files

chaitany6143:


src/components/SideNavBar/SideNavBarItemGroup.jsx
```diff
@@ -1,22 +1,40 @@
...
 import style from './style.scss';
 
-class SideNavBarItemGroup extends Component {
+
+type Props = {
+  id: string,
+  icon: string,
+  iconSize: number,
+  iconStyle?: Object,
+  label: Node | string,
```
kangax: https://flow.org/en/docs/react/types/#toc-react-node

Node could be a string already so this is redundant

src/components/SideNavBar/SideNavBarTop.jsx
```diff
@@ -1,10 +1,16 @@
...
-import PropTypes from 'prop-types';
 import cn from 'classnames';
 import Icon from '../Icon/Icon';
 import style from './style.scss';
 
-const SideBarNavTop = ({ label, onClick, icon }) => {
+type Props = {
+  label: string,
+  icon: string,
+  onClick: () => mixed,
```
kangax: shouldn't onClick always return void?

PR #196, opened 2018-10-26
fix typo

pseuyi:
`buttonType` spelled incorrectly within file


kangax:
@pseuyi should we get this in?

PR #197, opened 2018-11-05
change functionallity in componentWillReceiveProps so it can get an e…

shiranKfir:
Change functionality in componentWillReceiveProps so it can get an empty value.


pseuyi:
@shiranKfir has this been tested in spacestation?

kangax:
@shiranKfir we don't need package-lock.json since we're using yarn.lock already

PR #199, opened 2018-11-25
CC-134 - Minor improvements to Modal

GeReV:
Originally this task was to correct border-radius of Modal. Looks like someone already did that, so I merely improved the code slightly.

### DIFF
--- a/src/components/Modal/style.scss
+++ b/src/components/Modal/style.scss
@@ -3,7 +3,7 @@
 @keyframes slidein {
   from {
     opacity: .5;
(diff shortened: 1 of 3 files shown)


src/components/Modal/style.scss
```diff
@@ -74,11 +74,7 @@
 .pageActions {
   display: flex;
 
-  * {
-    margin-right: 20px;
-  }
-
-  :last-child {
-    margin-right: 0;
+  > * + * {
```
pseuyi: nice

PR #202, opened 2018-12-03
CC-155 - Fixed Button styling

GeReV:
Fixed height of Button to 40 pixels.

Also cleaned up the code a bit.


src/components/Button/Button.jsx
```diff
@@ -1,68 +1,80 @@
...
-    if (this.props.loading) {
+    if (loading) {
+      const loaderDotStyle = (type === Variants.SECONDARY || type === Variants.TERTIARY) ?
+        { backgroundColor: '#000', opacity: '0.1' } :
+        null;
+
       contentComponent = <Loader dotStyle={loaderDotStyle} />;
     } else {
-      contentComponent = this.props.label || this.props.children;
+      contentComponent = children || label;
```
GeReV: `children` should _always_ take precedent.

src/components/Button/Button.jsx
```diff
@@ -1,77 +1,80 @@
...
 
-    let contentComponent;
-    if (this.props.loading) {
-      contentComponent = <Loader dotStyle={loaderDotStyle} />;
-    } else {
-      contentComponent = this.props.label || this.props.children;
-    }
+    contentComponent = <Loader dotStyle={loaderDotStyle} />;
+  } else {
+    contentComponent = children || label;
```
GeReV: `children` should _always_ take precedent.

src/components/Button/Button.jsx
```diff
@@ -1,76 +1,86 @@
...
+
+import { getDataAttrs } from '../../dataUtils';
+
 import type { Data } from '../../types';
 
-const variants = { PRIMARY: 'primary', SECONDARY: 'secondary', TERTIARY: 'tertiary' };
-const size = { SMALL: 'small' };
+import style from './style.scss';
+
+const Variants = { PRIMARY: 'primary', SECONDARY: 'secondary', TERTIARY: 'tertiary' };
```
kangax: we gotta do that annoying `Object.freeze` wrapping here

src/components/Button/Button.jsx
```diff
@@ -1,76 +1,86 @@
...
+const Sizes = { SMALL: 'small' };
```
chelseawework: In the future, it might be a good idea to remove the "small" option. It is not in the Plasma design docs.

chelseawework:
According to `continuous-integration/travis-ci/p`, this is failing flow. It should pass before it is merged.

src/components/Button/Button.jsx
GeReV: I agree.

GeReV:
Right! Thanks you!

PR #203, opened 2018-12-05
CC-156 - Updated Label component

GeReV:
Gave the Label component a quick overhaul.

Now has _inline_ and _required_ states.
Added support for a few additional important attributes.

**Possible breaking change:** `children` props will now supersede `text` when both are passed.

Also added support for `className`, `id` and `name` attributes on `TextInput` (which deserves its own overhaul).


src/components/Label/Label.jsx
```diff
@@ -3,48 +3,52 @@ import React, { type Node } from 'react';
...
-const types = {
-  PRIMARY: 'primary',
-  SECONDARY: 'secondary',
-  DISABLED: 'disabled',
-};
+import styles from './style.scss';
 
 type Props = {|
-  text: string,
-  type: string,
```
chelseawework: The `type` isn't mentioned in Plasma design documentation. However, removing the `type` prop is breaking changes.

src/components/Label/Label.jsx
```diff
@@ -3,48 +3,52 @@ import React, { type Node } from 'react';
...
   children?: Node,
-  data?: Data,
   className?: string,
+  data?: Data,
+  disabled?: boolean,
+  htmlFor?: string,
```
chelseawework: Why do we need `htmlFor` prop? It is not mentioned in the plasma design documentation.

stories/Label.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Label from '../src/components/Label/Label';
+import TextInput from '../src/components/TextInput/TextInput';
+
+const stories = storiesOf('Label', module);
+
+stories.add('default', () => (
```
chelseawework: 🎉 Thanks for adding additional storybook views!

src/components/TextInput/TextInput.jsx
```diff
@@ -4,7 +4,7 @@ import React, { type Node } from 'react';
 import cx from 'classnames';
 import { getDataAttrs } from '../../dataUtils';
 import type { Data } from '../../types';
-import style from './style.scss';
+import styles from './style.scss';
```
chelseawework: 👍 I agree with changing `style` to `styles`. It makes more sense.

src/components/TextInput/TextInput.jsx
```diff
@@ -20,6 +20,9 @@ type Props = {|
   error: boolean,
   maxLength: string,
   data: Data,
+  id?: string,
+  name?: string,
+  style?: { [key: string]: any },
```
chelseawework: Why pass styles into this component directly? This is the component library the styles needed should already be included within it.

src/components/TextInput/TextInput.jsx
```diff
@@ -36,37 +39,42 @@ const TextInput = ({
   error,
   data,
   maxLength,
+  id,
+  name,
+  style,
```
chelseawework: Why did you include `TextInput` component in these changes?

src/components/Label/Label.jsx
GeReV: `htmlFor` is a `<label>`-specific attribute ( that is assigned the `id` of a related input tag.
When a label has this attribute (which translates to `<label for="...">`), clicking it sets focus to the relevant input, and also ties them together semantically.

Check out: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes

src/components/Label/Label.jsx
GeReV: Yep. The plan was to fix all the breaking changes in SS when upgrading its Plasma dependency.

src/components/TextInput/TextInput.jsx
GeReV: I did this for the sake of openness, but we can discuss if we want to keep this and/or `className`.
I'm in favor of at least leaving the latter.

src/components/TextInput/TextInput.jsx
GeReV: WDYM?

GeReV: Oh.
It's because it has to support `name` in general and `id` specifically for `htmlFor` to work.

PR #204, opened 2018-12-05
CC-158 - Improved TextInput component

GeReV:
Improved the interface for TextInput and rewrote some stuff.
Added a bunch of important attributes.

**BREAKING CHANGES:**
- `isDisabled` prop is no longer supported. Use `disabled`.
- Possible breaking: `suffix` and `prefix` are no longer mutually exclusive. `prefix` now appears _before_ the `input` in the DOM.


src/components/TextInput/TextInput.jsx
```diff
@@ -1,96 +1,77 @@
...
-  });
+class TextInput extends Component<Props> {
+  static defaultProps = {
+    placeholder: 'Type something...',
+  };
 
-  const wrapperStyle = cx(styles.wrapper, {
-    [styles.wrapperDisabled]: isDisabled || disabled,
-    [styles.wrapperError]: error,
-    [styles.wrapperLarge]: toUpper(size) === 'LARGE',
```
chelseawework: Since this was removed, it is a breaking change. This would need to be added back or addressed before we move to this version of Plasma.

GeReV: This will be addressed, and hopefully caught early by Flow where possible.

The design system doesn't address this style at all, from what I saw.

src/components/TextInput/style.scss
```diff
@@ -2,109 +2,79 @@
...
+  &:disabled {
+    @extend %input--disabled;
+  }
 }
 
-.fix {
-  @extend %fontFamily--body;
+.prefix,
+.suffix {
+  border: 0 solid $cGray300;
```
pseuyi: is this empty border needed?

src/decorators/withCounter/withCounter.jsx
```diff
@@ -1,26 +1,38 @@
...
 import cx from 'classnames';
-import style from './style.scss';
+
+import styles from './style.scss';
 
 type Props = {
   value: string,
-  counterStyle: any,
-  maxLength: string,
+  counterClassName?: string,
```
pseuyi: not sure if we need classname is a prop for this one

GeReV: I thought it might be necessary just in case, since we had `counterStyle` here.

src/components/TextInput/style.scss
GeReV: I'm actually just setting `border-width` and `border-color`, but IDEs sometimes complain that having those separately can be optimized.

If you'll look, I later set `border-right/left` for `.suffix/prefix`.

src/components/TextInput/TextInput.jsx
chelseawework: Since the property is removed, when this is merged into Plasma. Right after, the version would need to updated and any dependent repositories moved to this version and fixed (to account for this change). 

Otherwise, we would have a merged breaking change that would go unnoticed until someone else updated the Plasma version.

chelseawework: Since the property is removed, right after this is merged into Plasma, the version would need to updated and any dependent repositories moved to this version and fixed (to account for this change). 

Otherwise, we would have a merged breaking change that would go unnoticed until someone else updated the Plasma version.

PR #205, opened 2018-12-05
CC-157 - Added Hint component

GeReV:
Added the Plasma Hint component: http://plasma.guide/form-specs/#hint-text

This is a new component and therefore non-breaking.

screen shot 2018-12-05 at 18 24 35
screen shot 2018-12-05 at 18 24 42
screen shot 2018-12-05 at 18 24 46
screen shot 2018-12-05 at 18 24 50
screen shot 2018-12-05 at 18 24 57


src/components/Hint/Hint.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+import styles from './style.scss';
+
+type Props = {|
+  children?: Node,
+  className?: string,
+  data?: Data,
+  disabled?: boolean,
+  error?: boolean,
+  style?: { [key: string]: any },
```
chelseawework: I think this prop should be removed. If we can override the styles of the component itself, then what designs are we enforcing?

RainbowStripes (commented):
To help shed some light on these changes, could you add the Plasma style guide links (http://plasma.guide/form-specs/#hint-text) and pictures of the storybook to the PRs?

src/components/Hint/Hint.jsx
GeReV: We give a sane default, which is the right design. But I trust consumers to use this for things like layout or composition.
For example, if you're embedding it in a more complex component, you might need to address it from your CSS.

eladitzhakian (approved):
💯

setevik (approved):

PR #207, opened 2018-12-05
Add flow types generation, fix flow types

kangax:
Even though flow has `gen-flow-files` command, it's experimental and doesn't really work — https://github.com/facebook/flow/issues/5871

People are using various workarounds for generating lib types — https://blog.kentcdodds.com/distributing-flow-type-definitions-for-node-and-browser-modules-3952ad38b357

One of the solutions is to create `.flow` files that are simply a copy of plasma source files. This increases npm package size significantly, and that might have been fine except that it doesn't really work. We bundle plasma into a single file and Flow doesn't know which modules are exported and how to annotate them.

What I've decided to do for now is to automatically generate flow-typed defs in plasma and use that in SS. This is generally what a lot of libraries do (react-redux, lodash, etc.)

My solution has issues though. I'm extracting types from all the .jsx files under src/ using jscodeshift (same util that's used in codemods). The classes are exported and the types that those classes might use (e.g. `Tab`, `Item`) are declared locally.

<del>One of the issues is that references like `ExternalSelect.propTypes` are not currently resolved into proper types. For now, I replace them with `Object` but there might be a way to do it properly.</del> [update: now fixed]

<del>Similarly, we need to do something about `typeof some_var` — those need to be resolved properly and substituted.</del> [update: now fixed]

Finally, all the types and classes are exported in the same scope so there's a chance of conflicts. I'm not sure what the best solution for this is; we could scope-rename the types (e.g. `type Item` from `tabs.jsx` becomes `type tabs_Item`) and replace all occurrences in types.

- [x] normalize events
- [x] resolve `typeof ...` references

Update: we might be able to use `flow type-at-pos` to resolve types with all the values


scripts/generateFlowTypes.js
```diff
@@ -0,0 +1,30 @@
...
+  `jscodeshift -s -t scripts/extractFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
+);
+
+console.log(`
+// flow-typed version: <<STUB>>/@wework-dev/plasma_v${currentVersion}/flow_v${flowVersion}
+
+import type { ComponentType, ElementConfig, Node, Component } from 'react';
+
+declare module '@wework-dev/plasma' {
+  declare type Data = { key: string, value: string };
```
GeReV: Shouldn't this be a `{ [key: string]: string }`?

src/components/NumberInput/NumberInput.jsx
```diff
@@ -10,28 +10,28 @@ type HandleEvent = {|
...
+  },
+  nativeEvent: {
     target: {
-      value: Object
-    }
-  }
+      value: Object,
+    },
+  },
 |};
```
GeReV: Looks like a mock (and roundabout) definition of `SyntheticEvent<T>`

src/components/RadioButton/RadioButton.jsx
```diff
@@ -7,18 +7,17 @@ import { getDataAttrs } from '../../dataUtils';
...
+  checked?: boolean,
+  description?: ?string,
+  disabled?: ?boolean,
+  fancy?: ?boolean,
+  fieldValue?: string,
+  icon?: ?string,
+  isLarge?: ?boolean,
   name: string,
-  onChange: () => mixed,
+  onChange: (event: SyntheticEvent<HTMLInputElement>) => mixed,
```
GeReV: I suggest `evt`. I think `event` should be considered a saved word.

src/components/TimePicker/TimePicker.jsx
```diff
@@ -4,23 +4,23 @@ import style from './style.scss';
 import type { Data } from '../../types';
 
 type Props = {|
-  onBlur: (event: { target: {value: Object} }) => void,
-  onChange: (event: { target: {value: Object} }) => void,
+  onBlur?: (event: { target: { value: Object } }) => void,
+  onChange?: (event: { target: { value: Object } }) => void,
```
GeReV: `SyntheticEvent`s?

src/components/TimePicker/TimePicker.jsx
```diff
@@ -4,23 +4,23 @@ import style from './style.scss';
...
+  data?: Data,
+  disabled?: boolean,
 |};
 
 class TimePicker extends Component<Props> {
   static displayName = 'Plasma@TimePicker';
 
-  handleBlur = (event: { target: {value: Object} }): void => {
+  handleBlur = (event: { target: { value: Object } }): void => {
     const newTime = event.target.value;
```
GeReV: These should also be corrected to `event.currentTarget.value` for Flow support.

scripts/buildFlowTypeDefs.js
```diff
@@ -0,0 +1,29 @@
...
+
+declare module '@wework-dev/plasma' {
+  declare type Data = { key: string, value: string };
+
+  ${String(declarations)
+    .replace('<propTypes>', '<Object>')
+    .replace('<propTypes>', '<Object>')}
+
+  declare export class CountedTextInput extends React$Component<{ ...$PropertyType<TextInput, 'props'>, counterStyle: string }> { }
+  declare export class CountedTextArea extends React$Component<{ ...$PropertyType<TextArea, 'props'>, counterStyle: string }> { }
```
kangax: ideally we would generate decorated components automatically

scripts/buildFlowTypeDefs.js
```diff
@@ -0,0 +1,29 @@
...
+
+  declare export var LocationPin: string;
```
kangax: should we automatically export all files under icons as strings?

GeReV:
@kangax, sounds good with either approach.

scripts/buildFlowTypeDefs.js
```diff
@@ -12,7 +13,9 @@ const declarations = execSync(
 console.log(`
 // flow-typed version: <<STUB>>/@wework-dev/plasma_v${currentVersion}/flow_v${flowVersion}
 
-import type { ComponentType, ElementConfig, Node, Component } from 'react';
+import type { Node } from 'react';
+
+${String(declarations)}
```
GeReV: I think you want to move these two lines under `declare module`.

PR #208, opened 2018-12-06
[RFC] CC-161 - Helper Flow types

GeReV:
Added a bunch of commonly used Flow types for props.

Chosen types were based on:
- https://reactjs.org/docs/events.html
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes


src/types.js
```diff
@@ -3,3 +3,54 @@ export type Data = {
   key: string,
   value: string,
 };
+
+export type GlobalAttributes = $Shape<{|
+  hidden: boolean,
```
kangax: does every component accept these or what is this for exactly?

src/types.js
```diff
@@ -3,3 +3,54 @@ export type Data = {
...
+  id: string,
+  style: { [key: string]: any },
+  tabIndex: string | number,
+  title: string,
+|}>;
+
+// Event handlers
+
+export type SyntheticMouseEventHandler<T: EventTarget = EventTarget> =
```
kangax: these look good; we can then use them in SS as well. I'd like to see how/where exactly we could use them in SS and which advantages do they give us? will this fix `event.currentTarget`, etc. issues?

src/types.js
GeReV: Yes: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden.

> The `hidden` global attribute is a Boolean attribute indicating that the element is not yet, or is no longer, relevant.
> For example, it can be used to hide elements of the page that can't be used until the login process has been completed.

src/types.js
GeReV: It should, yes. I do think it could be helpful.
We should also maybe add an ESLint warning against using `evt.target` as well.

src/types.js
```diff
@@ -3,3 +3,63 @@ export type Data = {
   key: string,
   value: string,
 };
+
+export type GlobalAttributes = $Shape<{|
+  hidden: boolean,
+  id: string,
+  style: { [key: string]: any },
```
chelseawework: After our meeting: please replace `style` with `classNames` and make it optional.

GeReV: Aaaand done.

GeReV: By the way, since this is wrapped in `$Shape<>`, all these props are optional.

PR #212, opened 2018-12-17
[CMOS-4505] Update packages for React 16 compat

kangax:
Including master version of Plasma in SS currently breaks some pages. It looks like some of the Plasma dependencies are still pulling old version of React. This PR fixes it.


package.json
```diff
@@ -72,8 +72,8 @@
     "react-docgen-fixed": "^2.9.1",
     "react-dom": "16.6.3",
     "react-hot-loader": "^3.0.0-beta.6",
-    "react-router": "^2.6.0",
```
kangax: why do we even need react-router in plasma?

maniator: I am not sure we should

maniator: @kangax it is not used anywhere in this repository -- I believe it can be removed as a dep

kangax: removed

package.json
```diff
@@ -94,7 +93,7 @@
     "proxy-middleware": "^0.15.0",
     "react-datepicker": "^1.4.0",
```
maniator: shouldn't things like this be dependencies (or peed deps?)

kangax: what do you mean? they are under dependencies

package.json
```diff
@@ -94,7 +93,7 @@
...
     "react-input-autosize": "^1.1.0",
-    "react-select": "^0.9.1"
+    "react-select": "1.2.1"
```
yale: 😱

kangax: I checked in SS and it seems to look fine. @GeReV said API is the same in 1.x

PR #213, opened 2019-01-16
Add onKeyDown support in Search

kangax:
This is needed to be able to cancel events when typing

### DIFF
--- a/src/components/Search/Search.jsx
+++ b/src/components/Search/Search.jsx
@@ -25,6 +25,7 @@ type State = {|
 class Search extends Component<Props, State> {
   static defaultProps = {
     placeholder: 'Search...',
+    onKeyDown: null,
   };
 
   constructor(props: Props) {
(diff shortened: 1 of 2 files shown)


src/components/Search/Search.jsx
```diff
@@ -11,20 +11,23 @@ type Props = {|
   iconUrl: string,
   instructionText: string,
   onChange: ({target: {value: string}}) => void,
+  onKeyDown: (event: Event) => void,
```
aaroncouture: doesnt this need to be nullable?

aaroncouture: ie. onKeyDown?: (event: Event) => void

src/components/Search/Search.jsx
```diff
@@ -11,20 +11,23 @@ type Props = {|
...
 
 type State = {|
-    text: string
+  text: string
 |};
 
 class Search extends Component<Props, State> {
   static defaultProps = {
     placeholder: 'Search...',
+    onKeyDown: null,
```
aaroncouture: seems strange that we have to do this

kangax: this is because React doesn't like `undefined` in props, but asks for `null` instead

src/components/Search/Search.jsx
kangax: I'm guessing this works because we have it in default props

PR #214, opened 2019-01-18
Allow sidenav contents to scroll Y

yale:


src/components/SideNavBar/style.scss
```diff
@@ -51,6 +51,7 @@
   box-sizing: border-box;
   border-right: 1px solid $cBorder;
   padding: 20px 0 50px 0;
+  overflow-y: scroll;
```
maniator: weird that it did not do this by default

maniator (approved):
The lines changed seem off.

But LGTM

pseuyi (approved):

pseuyi:
removed `/dist`?

yale:
@pseuyi 	@maniator I ran `yarn run pack` and it replaced the CSS with a minified version... not sure why it hasn't always been minified tbh

maniator:
curious as to why we are even committing the dist folder

kangax (approved):
Kinda bizarre that we have SideNavBar in plasma...

PR #215, opened 2019-01-22
Update version script to build in prod mode

yale:
The README advises to run `yarn pack` when building a new version, which correctly runs `webpack --config webpack.config.prod -p`. 

However, those prod assets are overwritten when running `npm version patch`, which runs `webpack --config webpack.config.prod` during the `version` phase.

The `version` script is missing the `-p` flag, which is what signals to webpack to perform prod optimizations such as uglification and minification.

### DIFF
--- a/version.js
+++ b/version.js
@@ -15,7 +15,7 @@ fs.access('./dist/styles.css', fs.constants.R_OK, (err) => {
   }
 });
 
-const webpack = exec('webpack --config webpack.config.prod', (err, stdout, stderr) => {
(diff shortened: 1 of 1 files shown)


GeReV:
Just making sure, did you check it doesn't output CSS class names in that mode? They might be configured differently.

GeReV (approved):

yale:
@GeReV looks like it does output CSS class names. everything looks good on the branch that uses the new version (with the correctly prod minified assets): https://spacestation-v2-staging.wework.com/fix-sidenav-scroll/index.html

kangax (approved):

pseuyi (approved):

PR #218, opened 2019-02-04
CC-231 - Fixed Checkbox component

GeReV:
Gave `Checkbox` the same treatment that `TextInput` and `Label` others received.

This will be included in the 0.0.183 version of Plasma.

**Breaking changes:**
- `value` prop is now a string.
- Replaced `text` with `children`.
- Removed `fancy` prop. All checkboxes will look the same.
- Removed `description` prop. Use `Hint` component instead.
- Removed `indeterminate` prop (until we decide if it's necessary, since it can only be set via JS on the node).
- Removed `isBold` prop. Use `children` composition instead.
- Removed `icon` prop. Use `children` composition instead.
- Possibly breaking: Changed `onChange` and `onBlur` definitions to their canonical definitions.

Non-breaking changes:
- Added `error` boolean prop.
- Added `onClick`, `onFocus` props.
- Added [global attribute props](https://github.com/wework/plasma/blob/master/src/types.js#L7-L13).


setevik (approved):
Thanks!
You figured what was the actual visual issue with Chrome 72?
If it was SVG in `background-image`, we don't need to remove other 2 occurrences too?

src/components/Checkbox/Checkbox.jsx
```diff
@@ -1,92 +1,64 @@
 // @flow
-import { uniqueId } from 'lodash';
 import React, { type Node } from 'react';
+import { uniqueId } from 'lodash';
```
eladitzhakian: Interesting... 🧐

src/components/Checkbox/style.scss
```diff
@@ -1,151 +1,72 @@
...
-}
+  transition-duration: 0.1s;
+  transition-property: border-color, box-shadow, color;
+  transition-timing-function: ease-out;
 
-.textBold {
-  font-weight: bold;
-}
+  &:checked + .label::before {
+    background: url('data:image/svg+xml;utf8,<svg width="12px" height="10px" viewBox="0 0 12 10" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><title>checked</title><defs></defs><g id="Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd"><g id="checkbox-selected" transform="translate(-2.000000, -8.000000)" fill="#252729"><polygon id="checked" points="6 18 2 13.8333333 3.6 12.1666667 6 14.6666667 12.4 8 14 9.66666667"></polygon></g></g></svg>') center center no-repeat;
```
eladitzhakian: Can we get rid of this thing right here right now?

src/components/Checkbox/style.scss
```diff
@@ -1,151 +1,72 @@
...
+  }
 
-.textDisabled {
-  color: $cGray200;
-}
+  &:checked:disabled + .label::before {
+    background-image: url('data:image/svg+xml;utf8,<svg width="12px" height="10px" viewBox="0 0 12 10" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><title>checked</title><defs></defs><g id="Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd"><g id="checkbox-selected" transform="translate(-2.000000, -8.000000)" fill="#c2c2c2"><polygon id="checked" points="6 18 2 13.8333333 3.6 12.1666667 6 14.6666667 12.4 8 14 9.66666667"></polygon></g></g></svg>');
```
eladitzhakian: Same for this guy

eladitzhakian (approved):

src/components/Checkbox/style.scss
GeReV: Sadly no.

src/components/Checkbox/style.scss
GeReV: It's the same icon in a different color.

kangax:
I would really want us to hook up Plasma stories to Percy so that we can avoid any visual regressions.

GeReV:
Sounds reasonable. I'm not familiar with this repo's CI, though. Are we planning to replace it, so I could just copy from the SS repo?

PR #219, opened 2019-02-04
CC-213 - Fixed Checkbox SVG to work in Chrome 72

GeReV:
Chrome started being stricter about data URIs.


GeReV:
@setevik that is correct.

PR #221, opened 2019-02-19
Add new TimePicker Component

WeAvivOren:
Replace the TimePicker 
jira ticket - https://jira.we.co/browse/CMOP-351


src/components/TimePicker/TimePicker.jsx
```diff
@@ -1,46 +1,185 @@
...
-class TimePicker extends Component<Props> {
-  static displayName = 'Plasma@TimePicker';
+type State = {|
+  value: string,
+|};
 
-  handleBlur = (event: { target: { value: Object } }): void => {
-    const newTime = event.target.value;
+class TimePicker extends React.Component<Props, State> {
+  static defaultProps = {
```
marister: What happened eventually with figuring out how mandatory props in flow play together with default props?

src/components/TimePicker/TimePicker.jsx
```diff
@@ -1,46 +1,185 @@
...
+    minTime: '08:00',
```
marister: Defaults should probably from 00:00 to 23:59 or something like that. 8 to 20 is pretty random.

src/components/TimePicker/TimePicker.jsx
WeAvivOren: if a value is undefined it will be from the default props. flow doesn't recognize that.

src/components/TimePicker/TimePicker.jsx
WeAvivOren: we work office hours are 8-20

src/components/TimePicker/TimePicker.jsx
kangax: If you have defaultProps, Flow understands that it's optional

WeAvivOren: @kangax if i remove the ? optional for an attribute that is also in defaultProps, flow still shows the error of undefined.

src/components/TimePicker/TimePicker.jsx
tetraeder: It varies, think it would be better if it will be customable as well

WeAvivOren: its optional prop the default is 8-8, you can change it by passing minTime / maxTime

tetraeder: agree with @marister on that, it's a general time picker component, not coupled specifically to conference rooms

package.json
```diff
@@ -113,7 +113,8 @@
     "proxy-middleware": "^0.15.0",
     "react-datepicker": "^1.4.0",
     "react-input-autosize": "^1.1.0",
-    "react-select": "1.2.1"
+    "react-select": "1.2.1",
+    "moment": "^2.18.1"
```
GeReV: This should probably be added to the peer dependencies (https://nodejs.org/en/blog/npm/peer-dependencies/) rather than the dependencies themselves.
Am I right, @yale, @SergeyPanchenkoWeWork?

SergeyPanchenkoWeWork: Ye I think you right, but
1) We actually don't have moment in SS we have moment-timezone
2) There are many more changes required in package.json in Plasma (Separate issue)
3) If you are at it, probably should move lodash to peer dependency as well.
4) there already moment in plasma package.json devDependencies - so need to remove it from there as well

src/components/TimePicker/TimePicker.jsx
```diff
@@ -64,8 +63,8 @@ type Props = {|
 
 class TimePicker extends React.Component<Props> {
   static defaultProps = {
-    minTime: '08:00',
-    maxTime: '20:00',
+    minTime: '00:00',
+    maxTime: '24:00',
```
GeReV: I'm a bit concerned about allowing both "00:00" and "24:00".
It would work with Moment and get the correct representation, but I don't love it...

webpack.config.prod.js
```diff
@@ -29,7 +32,7 @@ module.exports = {
         loader: 'babel-loader',
       },
       {
-        test: /\.scss$/,
+        test: /\.(scss|css)$/,
```
GeReV: I recommend adding a different loader just for CSS, that won't include `sass-loader` in the `use` array below.

src/components/TimePicker/TimePicker.jsx
GeReV: I guess it's the right choice, though... it would still appear as "00:00" when printed, most likely...

GeReV:
I think you shouldn't change `dist/index.js` and `dist/style.css` in this PR.
We'll open a new branch that will add those changes as a new Plasma version.

src/components/TimePicker/TimePicker.jsx
WeAvivOren: its ok, the first and the last value are 12

package.json
```diff
@@ -82,7 +82,7 @@
     "lodash": "^4.14.1",
     "marked": "^0.3.6",
     "mkdirp": "^0.5.1",
-    "moment": "^2.18.1",
+    "moment": "2.24.0",
```
marister: We crossed test that didnt break anything?

GeReV: This specifically doesn't matter. We'll check this in the adaptation of SS.

PR #226, opened 2019-02-27
Plasma 0.0.188

GeReV:
- Exporting `Hint` component.
- **Breaking:** Updated the TimePicker component.


GeReV:
@setevik, will be in this one, but a bit later.

PR #228, opened 2019-03-07
Toggle - disabled items shouldn't call onChange()

kobim:
- Disabled items shouldn't call `onChange()`.
- Add `data-state` property to help test whether the item is enabled or disabled.

### DIFF
--- a/src/components/Toggle/Toggle.jsx
+++ b/src/components/Toggle/Toggle.jsx
@@ -33,7 +33,9 @@ class Toggle extends React.Component<Props> {
     onChange: () => {},
   };
   onClick(toggle: Object, index: number) {
-    this.props.onChange(toggle.label, index);
+    if (toggle.disabled !== true) {
+      this.props.onChange(toggle.label, index);
+    }
   }
 
   onKeyDown(event: Object, toggle: Object, index: number) {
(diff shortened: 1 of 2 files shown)


src/components/Toggle/Toggle.jsx
```diff
@@ -33,7 +33,9 @@ class Toggle extends React.Component<Props> {
     onChange: () => {},
   };
   onClick(toggle: Object, index: number) {
-    this.props.onChange(toggle.label, index);
+    if (toggle.disabled !== true) {
```
GeReV: I think you can write just `!toggle.disabled`.

What's `toggle`, by the way? Could you maybe change the annotation while you're here?

src/components/Toggle/Toggle.jsx
```diff
@@ -63,6 +65,7 @@ class Toggle extends React.Component<Props> {
               className={toggleClasses}
               key={index}
               type={this.props.type}
+              data-state={toggle.disabled ? 'disabled' : 'enabled'}
```
GeReV: Don't see a necessity for this, but OK.

kobim: there's no indication for _disabled_ items (expect of searching for "disabled" in the generated class name). This should help us testing.

PR #229, opened 2019-03-13
remove unused style prop and add innerWrapperStyle

shiranKfir:
- for our pre-registered guests bulk upload feature, I add to add a learn more link that flip the modal when clicked. In order for it to work, I add to add the ability to add style to the modal inner wrapper.

- I removed the style prop, it was unused.

here is a link to the pr in spacestation using the innerWrapperStyle prop: 
https://github.com/WeConnect/spacestation-web/pull/3873

### DIFF
--- a/src/components/Modal/Modal.jsx
+++ b/src/components/Modal/Modal.jsx
@@ -11,7 +11,7 @@ type Props = {|
   children: Node,
   visible?: boolean,
   onDismiss?: (evt: MouseEvent) => void,
-  style?: { [string]: string | number },
+  innerWrapperStyle: Object,
   minWidth?: string | number,
   maxWidth?: string | number,
   data?: Data,
@@ -28,6 +28,7 @@ class Modal extends React.Component<Props> {
       minWidth = 0,
       maxWidth,
       visible,
+      innerWrapperStyle,
     } = this.props;
 
     if (!visible) {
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.jsx
```diff
@@ -11,7 +11,7 @@ type Props = {|
   children: Node,
   visible?: boolean,
   onDismiss?: (evt: MouseEvent) => void,
-  style?: { [string]: string | number },
+  innerWrapperStyle: Object,
```
SergeyPanchenkoWeWork: I think  { [string]: string | number } is allitle bit better type

src/components/Modal/Modal.jsx
```diff
@@ -37,7 +38,7 @@ class Modal extends React.Component<Props> {
     return (
       <div {...getDataAttrs(data)} className={style.wrapper}>
         <div className={style.innerWrapper}>
-          <div style={{ minWidth, maxWidth }} className={style.card}>
+          <div style={{ minWidth, maxWidth, ...innerWrapperStyle }} className={style.card}>
```
SergeyPanchenkoWeWork: I would probably spread the object before `minWidth`  `maxWidth`

src/components/Modal/Modal.jsx
```diff
@@ -37,7 +38,7 @@ class Modal extends React.Component<Props> {
     return (
       <div {...getDataAttrs(data)} className={style.wrapper}>
         <div className={style.innerWrapper}>
```
SergeyPanchenkoWeWork: Why not put it the `innerWrapperStyle` on this wrapper? 
if you need in the card maybe better name would be `innerCardStyle`

PR #231, opened 2019-03-18
Time picker component

WeAvivOren:
Replace the TimePicker 
jira ticket - https://jira.we.co/browse/CMOP-351


GeReV:
@WeAvivOren, this is the latest code we had in SS, right?

PR #232, opened 2019-03-18
Plasma 0.0.191

GeReV:
Upgraded Plasma to 0.0.191.

- Upgraded `react-select` to 2.4.1 (therefore also `Select` component).
- Updated implementation of `TimePicker`.


GeReV:
@setevik doesn't really matter, I think. Just need to make sure they're present.

PR #234, opened 2019-03-21
Time picker component

WeAvivOren:


src/components/TimePicker/style.scss
```diff
@@ -44,7 +44,7 @@
 
     .time-picker__value-container {
       order: 1;
-      padding: 0;
+      padding: 1px;
```
GeReV: Really? Just 1px? that's odd.

src/components/TimePicker/TimePicker.jsx
```diff
@@ -226,20 +226,9 @@ class TimePicker extends React.Component<Props, State> {
...
-    const restProps = omit(
-      rest,
-      'defaultOption',
-      'hidden', // TODO(grozki): Get this to work?
-      'maxTime',
-      'minTime',
-      'timeFormat',
-      'timeIntervalMinutes',
-      'title' // TODO(grozki): Get this to work?
-    );
```
GeReV: I'm not sure you want to pass those, since React-Select doesn't handle those.
Might be worth to still omit them.

PR #236, opened 2019-04-24
added content style to modal

yotam707:
This PR will enable us to override the default style of the modal content for custom modal styling


src/components/Modal/Modal.jsx
```diff
@@ -15,6 +16,7 @@ type Props = {|
   minWidth?: string | number,
   maxWidth?: string | number,
   data?: Data,
+  contentStyle?: string,
```
GeReV: This should probably be `contentClassName`.

PR #238, opened 2019-05-14
add focused to search input

WeAvivOren:


src/components/Search/Search.jsx
```diff
@@ -78,6 +87,7 @@ class Search extends Component<Props, State> {
             placeholder={placeholder}
             type="text"
             value={this.state.text}
+            ref={input => (focused ? input && input.focus() : null)}
```
GeReV: Probably not the best solution. This will make it focus every time it renders.

Consider using the native [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus) instead.

stories/Search.js
```diff
@@ -1,7 +1,6 @@
+/* eslint-disable react/jsx-filename-extension */
```
GeReV: I think it's safe to rename this to `.jsx`.

PR #239, opened 2019-05-28
CMOS-5002 Export Button types to use in spacestation

power9:
Making this modification based on https://github.com/WeConnect/spacestation-web/pull/4763/files/e8b0064075c7d54391a7d4fcfca2633a22a39892#diff-e5a0efd3084db0b9dfe49166012cf21e

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@wework-dev/plasma",
-  "version": "0.0.193",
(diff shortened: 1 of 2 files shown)


kangax:
@power9 this is not going to work because we compile Plasma into one giant bundle and it has no types (you can see it in https://github.com/wework/plasma/blob/master/dist/index.js)

PR #240, opened 2019-06-10
FIX: dependencies warnings

kobim:
Fixes 2 warnings:

#### moment.js
```
warning package.json: "dependencies" has dependency "moment" with range "^2.18.1" that collides with a dependency in "devDependencies" of the same name with version "2.24.0"
warning @wework-dev/plasma@0.0.193: "dependencies" has dependency "moment" with range "^2.18.1" that collides with a dependency in "devDependencies" of the same name with version "2.24.0"
```

#### react-tap-event-plugin
Removed, as it is [deprecated](https://github.com/zilverline/react-tap-event-plugin#deprecated) and not in use by any of the dependencies or plasma itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "lodash": "^4.14.1",
     "marked": "^0.3.6",
     "mkdirp": "^0.5.1",
-    "moment": "2.24.0",
     "node-sass": "^4.9.3",
     "prettier": "1.16.1",
     "proxy-middleware": "^0.15.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -114,13 +113,12 @@
     "react-datepicker": "^1.4.0",
     "react-input-autosize": "^1.1.0",
     "react-select": "^2.4.1",
-    "moment": "^2.18.1"
```
kangax: I see we're using the same version in SS which is good but I wonder if having both moment-timezone (in SS) and moment (here) is blowing up our bundle or if webpack knows how to do this right?

kangax:
OT: we should probably move Plasma to Circle for consistency

package.json
kobim: it resolves to the same one (_yarn.lock_ of SS):
```
moment@2.24.0, "moment@>= 2.9.0", moment@^2.18.1, moment@^2.23.0, moment@^2.24.0:
  version "2.24.0"
```

kangax: great

PR #243, opened 2019-06-20
Allow Loader to accept classNames or dotColor

kobim:
Adding 3 new properties to `<Loader />`:
- `className` & `dotClassName` which will apply on the wrapper and the dot
```css
.coolDot {
  background-color: #00ff00 !important;
}
```
```jsx
  import styles from './styles.css';

  <Loader dotClassName={styles.coolDot} />
```

- `dotColor` for specifically overriding the dot color.
```jsx
  <Loader dotColor="#c1c1c1" />
```

The `import/no-extraneous-dependencies` rule was set to ignore the `storybook` import on the stories.

### DIFF
--- a/src/components/Loader/Loader.jsx
+++ b/src/components/Loader/Loader.jsx
@@ -6,19 +6,19 @@ import { getDataAttrs } from '../../dataUtils';
 import type { Data } from '../../types';
 
 type Props = {|
-  className?: Object,
+  className?: string,
   dotColor?: string,
-  dotClassName?: Object,
+  dotClassName?: string,
   style?: Object,
-  dotStyle?: ?Object,
+  dotStyle?: Object,
   data?: Data,
 |};
 
 class Loader extends React.Component<Props> {
   render() {
-    const dotStyle = this.props.dotStyle || {};
-    if (this.props.dotColor != null) {
-      dotStyle.backgroundColor = this.props.dotColor;
+    const { dotColor, dotStyle = {} } = this.props;
+    if (dotColor != null) {
+      dotStyle.backgroundColor = dotColor;
     }
 
     return (
--- a/src/components/Loader/Loader.jsx
+++ b/src/components/Loader/Loader.jsx
@@ -10,13 +10,17 @@ type Props = {|
   dotColor?: string,
   dotClassName?: string,
   style?: Object,
-  dotStyle?: Object,
+  dotStyle: Object,
   data?: Data,
 |};
 
 class Loader extends React.Component<Props> {
+  static defaultProps = {
+    dotStyle: {},
+  };
+
   render() {
-    const { dotColor, dotStyle = {} } = this.props;
+    const { dotColor, dotStyle } = this.props;
     if (dotColor != null) {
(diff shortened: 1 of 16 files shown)


src/components/Loader/Loader.jsx
```diff
@@ -1,23 +1,36 @@
...
+  dotClassName?: Object,
   style?: Object,
   dotStyle?: ?Object,
   data?: Data,
 |};
 
 class Loader extends React.Component<Props> {
   render() {
+    const dotStyle = this.props.dotStyle || {};
+    if (this.props.dotColor != null) {
```
GeReV: Maybe destructure some stuff?

src/components/Loader/Loader.jsx
```diff
@@ -1,23 +1,36 @@
...
 import React from 'react';
+import cn from 'classnames';
 import style from './style.scss';
 import { getDataAttrs } from '../../dataUtils';
 import type { Data } from '../../types';
 
 type Props = {|
+  className?: Object,
+  dotColor?: string,
+  dotClassName?: Object,
```
GeReV: class names should probably be strings.

src/components/Loader/Loader.jsx
```diff
@@ -1,23 +1,36 @@
...
   style?: Object,
   dotStyle?: ?Object,
   data?: Data,
 |};
 
 class Loader extends React.Component<Props> {
   render() {
+    const dotStyle = this.props.dotStyle || {};
```
kangax: make this a default prop?

stories/style.scss
```diff
@@ -0,0 +1,3 @@
+.greenDotLoader {
+  background-color: #00ff00 !important;
```
kangax: nit: `#0f0`

src/components/Loader/Loader.jsx
kobim: WDYT about https://github.com/wework/plasma/pull/243/files#diff-529cbf4c11cef5ef490162beb870bc08R19 ?

kangax: that's fine too, I was just thinking to follow a more idiomatic-reacty way to define a default value for a prop

kobim: my bad, I kept defining `dotStyle?: Object` when I tried the `defaultProps`. I forgot that `dotStyle: Object` is enough.

src/components/Loader/Loader.jsx
```diff
@@ -1,23 +1,40 @@
...
 import { getDataAttrs } from '../../dataUtils';
 import type { Data } from '../../types';
 
 type Props = {|
+  className?: string,
+  dotColor?: string,
+  dotClassName?: string,
   style?: Object,
-  dotStyle?: ?Object,
+  dotStyle: Object,
```
kangax: can we use `{ [string]: string }` here since `Object` is generally unsafer?

kobim: why not `StyleAttributes`? :)

kangax: perfect!

PR #244, opened 2019-07-11
Update package to be smaller on npm

maniator:
This reduces the bundle size significantly

From bundlephobia currently:
image

To this new size:
image


maniator:
> Datepicker still looks good?

@kangax we don't use the datepicker from plasma, so why would it look any different?

maniator:
> Since the `package.json` file has changes, this should be tested locally with current version of spacestation to ensure it doesn't break anything -just to be on the safe side.

I did that :-)

PR #248, opened 2019-07-12
Remove react-datepicker package that's unused for the past 2 years, lol

kangax:


src/external/index.js
```diff
@@ -1,9 +0,0 @@
-// import reactselect from './reactselect.scss';
-import reactdatepicker from './reactdatepicker.scss';
-
-const style = {
-  // reactselect,
-  reactdatepicker,
-};
```
kangax: need to check that we're not using any of this in SS

kangax: Checked, unused

maniator: All of the datepickers look kosher in ss?

docs/gitbook/setup/README.md
```diff
@@ -5,8 +5,7 @@ To get up and running developing with Plasma, follow these instructions:
 #### Important note
 
 ````
-Select and Datepicker components do not work right now.
-Use react-select and datepicker directly in Spacestation.
+Select component does not work right now. Use react-select directly in Spacestation.
```
maniator: Can we remove the select code as it is not used anywhere (and all commented out where it is used)

PR #250, opened 2019-07-15
CMOS-5144 Prepare plasma to be tree-shakeable

maniator:
This is a preparation to make plasma tree-shakeable by removing the "built" css file for plasma and using style loader instead to load styles for the library

This allows for us to split up the code to be easily tree shaken in the next follow-up


kangax (approved):
Let's get approvals from other teams coz it's kind of a big change

maniator:
> Let's get approvals from other teams coz it's kind of a big change

I feel like this will (finally) be a plasma minor update.
This will be followed up with tree shaking.

PR #251, opened 2019-07-16
CMOS-5144 Code split the plasma build to be tree-shakeable

maniator:
Basically created a new webpack config to create the dist output for plasma, and updated the prod config to output a file for every component/file within source so that everything can be tree-shaken

plasma is now tree-shakeable

https://bundlephobia.com/result?p=@wework-dev/plasma@0.1.1

vs

https://bundlephobia.com/result?p=@wework-dev/plasma@0.0.203


lib/components/Card/Card.js
```diff
@@ -0,0 +1,413 @@
+!(function(e, t) {
```
kangax: why are these files minified partially? e.g. names are all minified but the whitespace is not

maniator: lol oops. that was my IDE. I was inspecting this file. I can rerun

maniator: shouldnt matter in the long run. really the lib files should really not be minified (I could not figure out how to make webpack _not_ minify them)

.babelrc
```diff
@@ -1,9 +1,15 @@
...
+      "modules": "commonjs",
+      "useBuiltIns": true
+    }],
+    "react",
+    "flow"
+  ],
   "plugins": [
   	"transform-class-properties",
-  	"transform-object-rest-spread",
-  	"lodash"
```
maniator: Not needed anymore as lodash is an external dependency in webpack now.

GeReV: But that's the babel-lodash-plugin, it doesn't have anything to do with dependencies.

webpack.config.dev.js
```diff
@@ -30,24 +25,20 @@ module.exports = {
...
-        loader: ExtractTextPlugin.extract({
-          fallback: 'style-loader',
-          use: [
-            'css-loader?importLoaders=1&localIdentName=[path][local]__[hash:base64:5]',
-            'resolve-url-loader',
-          ],
-        }),
+        use: [
+          'style-loader',
+          `css-loader?importLoaders=1&localIdentName=${css.localIdentName}`,
```
GeReV: Split this into an object with options?
```
{
  loader: 'css-loader',
  options: {
    importLoaders: 1,
    localIdentName: css.localIdentName,
  }
}
```

webpack.config.dev.js
```diff
@@ -30,24 +25,20 @@ module.exports = {
...
-          fallback: 'style-loader',
-          use: [
-            'css-loader?modules&importLoaders=2&localIdentName=[path][local]__[hash:base64:5]',
-            'resolve-url-loader',
-            'sass-loader',
-          ],
-        }),
+        use: [
+          'style-loader',
+          `css-loader?modules&importLoaders=1&localIdentName=${css.localIdentName}`,
```
GeReV: You can also split this one.

webpack.config.dist.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable import/no-extraneous-dependencies, no-console */
+
+const css = require('./css.config');
+const path = require('path');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
```
GeReV: Maybe transition to `MiniCssExtractPlugin` if possible?

webpack.config.prod.js
```diff
@@ -16,10 +58,10 @@ module.exports = {
...
   },
   output: {
-    path: path.resolve(__dirname, 'dist'),
-    filename: 'index.js',
-    library: 'Plasma',
-    libraryTarget: 'umd',
+    path: path.resolve(__dirname, 'lib'),
+    filename: '[name].js',
+    library: '',
+    libraryTarget: 'commonjs2',
```
GeReV: So two questions:
1. Why do we have a different config for `prod` and `dist`? Why the different name?
1. Why does one use `umd` and the other `commonjs2`?

webpack.config.dist.js
maniator: This is using webpack 2 (which I just upped to 3)

I dont want to make too many changes just yet

webpack.config.prod.js
maniator: One is for a distributable file and the other is for a code split tree shaken file.

Eventually, when we remove react-select as a real dependency and don't need to bundle it in. we can just sue `babel src -d lib` and we will not need webpack at all for the tree shaken version.

webpack.config.dev.js
maniator: This is basically how it was before, I just wanted a way to change them all at the same time easily (in webpack 3 I dont even think you can do what you say here, you need to use `LoaderOptions` plugin which is a headache)

webpack.config.dist.js
GeReV: Fair enough.

webpack.config.prod.js
GeReV: Seems like `umd` covers `commonjs`, so why both?
How about `esm` (if supported)?

maniator: Possibly.

This is what works right now. 

The dist file still can only be *one* file, the module can be multiple using require, so that is what we are doing here -- we are using commonjs in spacestation, there is no need for umd

maniator: @GeReV we can get to esm and regular babel building when we remove react-select@2 as a direct dependency

.babelrc
maniator: It does. it converts all of the `lodash` imports to `lodash/<file>` which then screws with the externals in the webpack config and then builds locash into each entry (which we do not want... :-( )

webpack.config.prod.js
```diff
@@ -1,12 +1,56 @@
...
+  {
+    dataUtils: './src/dataUtils.js',
+  },
+  componentEntries,
+  decoratorEntries,
+  externalEntries,
+  iconEntries
+);
+
+console.log('ENTRIES', entries);
```
pseuyi: intentional?

css.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  localIdentName: '[path][local]__[hash:base64:5]',
```
pseuyi: is this..causing snapshot updates in the other pr?

maniator: Nope.

webpack.config.prod.js
maniator: Yes

css.config.js
maniator: it is lodash that is causing the diffs :-(

PR #253, opened 2019-07-16
Toggle: Allow styling

kobim:
Adding the option to provide a className and styling options for the various states (active, inactive and disabled):
Screen Shot 2019-07-16 at 21 34 54

### DIFF
--- a/stories/style.scss
+++ b/stories/style.scss
@@ -1,3 +1,16 @@
 .greenDotLoader {
   background-color: #0f0 !important;
 }
+
+.toggle {
+  &[data-active="true"] {
+    background-color: #ff01ff !important;
+    border-color: #9e1c9e !important;
+  }
+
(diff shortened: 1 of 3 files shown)


stories/style.scss
```diff
@@ -1,3 +1,16 @@
 .greenDotLoader {
   background-color: #0f0 !important;
 }
+
+.toggle {
+  &[data-active="true"] {
```
kangax: My only concern here is that we're exposing an implementation detail so any changes to data-... attributes could break things "upstream" in the consumers of these components. This might not necessarily be a bad thing if we consider data-* attributes (or all attributes) to be part of a public API. It's just that right now, afaik, the only API is component props and everything else (e.g. hooking onto classes or a specific structure of a component) should not be counted on.

PR #257, opened 2019-07-30
Table: Allow to specify the empty element

kobim:
Will allow to pass a JSX element in addition to a string to `emptyText`
Screen Shot 2019-07-30 at 13 49 16

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -12,8 +12,7 @@ import style from './style.scss';
 type Props = {|
   clickable?: boolean,
   empty: boolean,
-  emptyText: string,
-  emptyPlaceholder: ?React$Node,
+  emptyText: React$Node,
   headerData: Array<Object>,
   highlightable?: boolean,
   items: ?Array<Object>,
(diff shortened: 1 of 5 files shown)


kobim:
@kangax our issue is with the type of that placeholder (which can be only `string` with pre-defined style)
I've committed a change that might be more suitable - being able to pass `React$Node` and insert an Element with whatever style we want. (also had to give some love to flow and eslint in the process)

src/components/Table/Table.jsx
```diff
@@ -41,9 +41,7 @@ class Table extends React.Component<Props, State> {
     empty: false,
     emptyText: '',
     loading: false,
-    stickAt: null,
     headerData: [],
-    sort: {},
```
kangax: are these default values not used?

src/components/Table/Table.jsx
```diff
@@ -12,7 +12,7 @@ import style from './style.scss';
 type Props = {|
   clickable?: boolean,
   empty: boolean,
-  emptyText: string,
+  emptyText: React$Node,
```
kangax: Much better, thanks! `React$Node` also already includes a string so this just expands the type and shouldn't affect public API in any way.

src/components/Table/Table.jsx
kobim: it was always protected by `sort && ...`, and it caused flow issues with the tests (where it was annotated as a required attribute)

PR #258, opened 2019-08-07
Update Checkbox and Radios to only generate IDs on creation and not o…

maniator:
…n every render

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@wework-dev/plasma",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "description": "",
   "main": "dist/index.js",
   "module": "lib/index.js",
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -1,5 +1,4 @@
 // @flow
-import { uniqueId } from 'lodash';
 import React, { type Node } from 'react';
 import cx from 'classnames';
 import style from './style.scss';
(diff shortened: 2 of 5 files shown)


src/components/Checkbox/Checkbox.jsx
```diff
@@ -32,6 +33,8 @@ class Checkbox extends React.Component<Props> {
     value: false,
   };
 
+  id = this.props.id || uniqueId('id');
```
pseuyi: what's the reason to derive it from props?

maniator: in case someone wants to pass in an id (as people usually do... regular inputs support ids)

maniator:
> YES!
> but what about props changing?

@kobim the id on an input should not be changing... also, this should not really affect it at all..

kobim:
also wondering - why do we need `id` at all? can't we just use [implicit label association](https://www.w3.org/TR/html4/interact/forms.html#h-17.9.1) ? (we already nesting the `input` beneath the `label`)

maniator:
Ha I have no idea kobi, I just want to make it so that the current version does not Regen on every render.

We can do that later if we want to

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@wework-dev/plasma",
-  "version": "0.1.2",
+  "version": "0.1.3",
```
kangax: do we include version bump in feature/bug PR's now?

PR #260, opened 2019-08-19
Update flow types to better match what is actually being used

maniator:
also committing the flow typed defs for plasma

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,19 +19,19 @@
     "styleguide": "styleguidist server",
     "styleguide:build": "styleguidist build",
     "watch": "webpack --watch --config webpack.config.dev",
-    "preversion": "yarn build",
-    "prepublish": "yarn build",
+    "preversion": "yarn build && yarn flow:buildTypeDefs",
+    "prepublish": "yarn preversion",
     "test": "node_modules/.bin/jest",
     "test:watch": "yarn test --watch --coverage",
     "test:update": "yarn test --updateSnapshot",
     "test:coverage": "yarn test --coverage --coverageReporters=text-lcov",
     "flow:coverage": "flow-coverage-report --config .flow-coverage.config.json --strict-coverage",
-    "flow:buildTypeDefs": "node ./scripts/buildFlowTypeDefs.js"
+    "flow:buildTypeDefs": "node ./scripts/buildFlowTypeDefs.js > flow/plasma_vx.x.x.js"
   },
   "lint-staged": {
     "*.{js,jsx}": [
       "prettier --write",
-      "eslint",
+      "eslint --fix",
       "git add"
     ]
   },
--- a/scripts/buildFlowTypeDefs.js
+++ b/scripts/buildFlowTypeDefs.js
@@ -10,7 +10,7 @@ const defaultDeclarations = execSync(
   `jscodeshift -s -t scripts/extractDefaultFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
 );
 
-const componentDecclarations = execSync(
+const componentDeclarations = execSync(
   `jscodeshift -s -t scripts/extractComponentFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
 );
 
--- a/scripts/buildFlowTypeDefs.js
+++ b/scripts/buildFlowTypeDefs.js
@@ -6,8 +6,12 @@ const p = require('../package.json');
 const currentVersion = p.version;
 const flowVersion = String(execSync('flow version')).match(/[\d\.]+/)[0];
 
-const declarations = execSync(
-  `jscodeshift -s -t scripts/extractFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
+const defaultDeclarations = execSync(
+  `jscodeshift -s -t scripts/extractDefaultFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
+);
+
+const componentDecclarations = execSync(
+  `jscodeshift -s -t scripts/extractComponentFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
 );
 
 console.log(`
@@ -16,11 +20,13 @@ console.log(`
 declare module '@wework-dev/plasma' {
   declare type Data = { [key: string]: string };
 
-  ${String(declarations).replace(/\bNode\b/g, 'React$Node')}
+  ${String(defaultDeclarations)}
 
   declare export class CountedTextInput extends React$Component<{ ...$PropertyType<TextInput, 'props'>, counterClassName?: string }> { }
   declare export class CountedTextArea extends React$Component<{ ...$PropertyType<TextArea, 'props'>, counterClassName?: string }> { }
 
   declare export var LocationPin: string;
 }
+
+${String(componentDeclarations)}
 `);
--- a/src/components/SegmentedCard/SegmentedCard.jsx
+++ b/src/components/SegmentedCard/SegmentedCard.jsx
@@ -1,12 +1,12 @@
 // @flow
-import React, { type Node } from 'react';
+import React from 'react';
 import cx from 'classnames';
 import style from './style.scss';
 import { getDataAttrs } from '../../dataUtils';
 import type { Data } from '../../types';
 
 type Props = {|
-  children: Array<Node>,
+  children: React$Node,
   style: Object,
   vertical: Boolean,
   data?: Data,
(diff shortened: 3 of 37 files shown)


src/decorators/withCounter/withCounter.jsx
```diff
@@ -10,7 +10,7 @@ type Props = {
   maxLength?: string,
 };
 
-const withCounter = <P: {}>(Component: ComponentType<P>) => {
+const withCounter = <P: any>(Component: ComponentType<P>) => {
```
kangax: how did you end up fixing this?

stories/TextInput.jsx
```diff
@@ -1,4 +1,3 @@
-// @flow
```
kangax: why remove?

src/components/layout/Group/Group.jsx
```diff
@@ -40,7 +40,7 @@ class Group extends React.Component<Props> {
           } else if (this.props.layout[index]) {
             styleObj.flexBasis = this.props.layout[index];
           } else {
-            styleObj.flexBasis = `${100 / this.props.children.length}%`;
+            styleObj.flexBasis = `${100 / React.Children.count(this.props.children)}%`;
```
kangax: interesting, why this change?

src/components/layout/FixedLeft/FixedLeft.jsx
```diff
@@ -74,6 +75,8 @@ class FixedLeft extends React.Component<Props, State> {
   fixedViewportOffsetOrigin: number;
 
   render() {
+    const children = React.Children.toArray(this.props.children);
```
kangax: what's happening here?

scripts/buildFlowTypeDefs.js
```diff
@@ -6,21 +6,28 @@ const p = require('../package.json');
...
-const declarations = execSync(
-  `jscodeshift -s -t scripts/extractFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
+const defaultDeclarations = execSync(
+  `jscodeshift -s -t scripts/extractDefaultFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
 );
 
+const componentDeclarations = execSync(
+  `jscodeshift -s -t scripts/extractComponentFlowTypes.js src/components/**/*.jsx src/components/layout/**/*.jsx --parser flow`
+);
+const typeDec = execSync(`cat src/types.js`);
```
kangax: `fs.readFileSync`?

package.json
```diff
@@ -19,25 +19,25 @@
...
-    "prepublish": "yarn build",
+    "preversion": "yarn build && yarn flow:buildTypeDefs",
+    "prepublish": "yarn preversion",
     "test": "node_modules/.bin/jest",
     "test:watch": "yarn test --watch --coverage",
     "test:update": "yarn test --updateSnapshot",
     "test:coverage": "yarn test --coverage --coverageReporters=text-lcov",
     "flow:coverage": "flow-coverage-report --config .flow-coverage.config.json --strict-coverage",
-    "flow:buildTypeDefs": "node ./scripts/buildFlowTypeDefs.js"
+    "flow:buildTypeDefs": "node ./scripts/buildFlowTypeDefs.js > flow/plasma_vx.x.x.js"
```
kangax: we need to figure out how to "version" typedefs properly so that updating plasma in SS would yell if we also don't upgrade the types

src/decorators/withCounter/withCounter.jsx
GeReV: I think there's no need to specify the `: any` then.

src/components/layout/Group/Group.jsx
GeReV: That's actually more correct.

src/components/layout/FixedLeft/FixedLeft.jsx
GeReV: It's correct, according to the docs.

src/components/SegmentedCard/SegmentedCard.jsx
```diff
@@ -21,11 +21,12 @@ class SegmentedCard extends React.Component<Props> {
...
       [style.vertical]: this.props.vertical,
     });
+    const children = React.Children.toArray(this.props.children);
 
     return (
       <div {...getDataAttrs(this.props.data)} className={className} style={this.props.style}>
-        <div className={style.first}>{this.props.children[0]}</div>
-        <div className={style.second}>{this.props.children[1]}</div>
+        <div className={style.first}>{children[0]}</div>
+        <div className={style.second}>{children[1]}</div>
```
GeReV: You can't really ensure exactly two children, though...

src/components/Table/Table.jsx
```diff
@@ -23,7 +23,7 @@ type Props = {|
   stickAt?: number,
   style?: StyleAttributes,
   data?: Data,
-  sort?: { key: string, order: string },
+  sort?: { key: ?string, order: ?string },
```
GeReV: This seems wrong. If you specify a sort order it should at least include these two values (or at least the key with a default of ASC order).

src/components/Tabs/Tabs.jsx
```diff
@@ -8,12 +8,12 @@ import style from './style.scss';
...
 type Tab = {|
   label: string,
-  title: Node,
+  title: React$Node,
 |};
 
 type Props = {|
   items: Array<Tab>,
-  onChange?: (Node, number) => void,
+  onChange?: (string, number) => void,
```
GeReV: Tab title is a `React$Node`, though, no?

src/components/layout/Group/Group.jsx
maniator: This is a flow fix

As React$Node is not an array

src/components/Tabs/Tabs.jsx
maniator: Nope

src/components/TimePicker/TimePicker.jsx
```diff
@@ -132,19 +132,19 @@ class TimePicker extends React.Component<Props, State> {
   };
 
   static getDerivedStateFromProps = ({
-    maxTime,
-    minTime,
+    maxTime = TimePicker.defaultProps.maxTime,
+    minTime = TimePicker.defaultProps.minTime,
```
GeReV: Isn't this taken care of automatically?

src/components/SegmentedCard/SegmentedCard.jsx
maniator: True. But I did not change from the old code

src/components/TimePicker/TimePicker.jsx
maniator: Nope... Flow was complaining about it...

src/components/layout/FixedLeft/FixedLeft.jsx
```diff
@@ -26,8 +26,8 @@ type State = {|
 |};
 
 type Props = {|
-  children: Array<Node>,
-  stickAt: number,
+  children: React$Node,
```
GeReV: I think `children` can be optional, in general.

stories/TextInput.jsx
maniator: Because I could not figure out the issue.

Other places that used with counter worked fine (and I think that might be because the storybook file was not flowified, or it could be something else)

src/decorators/withCounter/withCounter.jsx
GeReV:
```suggestion
const withCounter = <P>(Component: ComponentType<P>) => {
```

maniator: Nope could not figure it out juriy

scripts/buildFlowTypeDefs.js
maniator: Same thing in the end, but true

package.json
maniator: It has the version at the top of the flow def file, so maybe we could use that

src/components/layout/FixedLeft/FixedLeft.jsx
maniator: I did not want to change tons of API in this PR only what was needed

src/components/Tabs/Tabs.jsx
maniator: Well maybe, but if you look at the code this function gets passed a string

src/components/Table/Table.jsx
maniator: Maybe... There was a flow issue in ss so this seemed to fix it

src/components/layout/FixedLeft/FixedLeft.jsx
maniator: Same as above comment: https://github.com/wework/plasma/pull/260#discussion_r316150721

scripts/buildFlowTypeDefs.js
maniator: Done.

PR #265, opened 2019-09-04
SPST-79: Consistent flow typedefs

kobim:
We want to avoid major file changes in the flow typedefs (#260 compared to #263).

jscodehint's worker [can't run in order](https://github.com/facebook/jscodeshift/blob/master/src/Worker.js#L132). Even if `runInBand` option is set, the worker still receives the whole list of files and processes them asynchronically.
To make it run in a sorted order, we will only pass 1 file at a time and use the `report()` function instead of `console.log`.


kobim:
> So this guarantees the same order of files in the flow typedefs file?

YES!

kangax:
@kobim nice! ship it

flow/plasma_vx.x.x.js
```diff
@@ -211,112 +109,98 @@ declare module '@wework-dev/plasma' {
...
-    type?: 'text' | 'password' | 'email' | 'url',
-    suffix?: React$Node,
-    prefix?: React$Node,
-    placeholder?: string,
-    onClick?: SyntheticMouseEventHandler<HTMLInputElement>,
-    name?: string,
-    maxLength?: number | string,
-    error?: boolean,
-    disabled?: boolean,
+  // source file: src/components/CollapsibleCard/CollapsibleCard.jsx
```
maniator: I guess we never noticed before that some things were twice (and sometimes empty)

kobim: you're right, let's remove them :)
69ed9d5cf0da8963bef48f1a26f84dbd241c612f

PR #271, opened 2019-09-19
Fix inexact in plasma flow types

maniator:
Updated version of eslint to support the new flow `...` inexact definition

Also updated the export flow-typed definitions to include the defs for all static variables and methos defined on all classes exported by plasma.

ESLint was updated to understand the new `...` syntax (together with fixing all the new lint errors)


src/components/TextInput/TextInput.jsx
```diff
@@ -4,7 +4,7 @@ import cx from 'classnames';
 
 import { getDataAttrs } from '../../dataUtils';
 import type {
-  ChangeEventHanlders,
+  ChangeEventHandlers,
```
kangax: lol

flow/plasma_vx.x.x.js
```diff
@@ -222,24 +220,26 @@ declare module '@wework-dev/plasma' {
...
-  declare type HandleEvent = {|
-    nativeEvent: { target: { value: any } },
-    target: { parentElement: { classList: { add: string => mixed, remove: string => mixed } } },
-  |};
+  declare type HandleEvent =
+    | {|
+        nativeEvent: { target: { value: any } },
+        target: { parentElement: { classList: { add: string => mixed, remove: string => mixed } } },
+      |}
+    | string;
```
kangax: is this correct? when would HandleEvent be a string?

src/components/NumberInput/NumberInput.jsx
```diff
@@ -20,19 +20,19 @@ type HandleEvent = {|
       value: Object,
     },
   },
-|};
+|} | string;
```
kobim: I'm with @kangax here too.. we should understand where this is being used with a string.

src/types.js
```diff
@@ -1,7 +1,5 @@
 // @flow
-export type Data = {
-  [key: string]: string,
-};
+export type Data = { [key: string]: string, ... };
```
kobim: I think this should be an exact type

src/types.js
```diff
@@ -67,4 +65,4 @@ export type FocusEventHandlers<T: EventTarget = EventTarget> = {|
   onBlur?: SyntheticFocusEventHandler<T>,
 |};
 
-export type StyleAttributes = { [string]: string | number };
+export type StyleAttributes = { [string]: string | number, ... };
```
kobim: maybe `{ [key: string]: string | number | { ... } }`? (as `StyleAttributes` will always have string keys.)

flow/plasma_vx.x.x.js
maniator: We are passing it a string in one place in plasma it seems

I can look more tomorrow when I am by my computer

src/types.js
kobim: ping ^^

kangax:
@kobim we need to tell Flow about React components' default props as it influences the way Flow sees props internally (used in a component) vs. externally (when component is used somewhere else) — https://flow.org/en/docs/react/components/#toc-using-default-props

Basically, default prop in Flow is marked as a non-optional property and it knows to not require it when the component is used somewhere else AND also not error out when you do smth like `this.props.propThatHasDefaultValue()` inside the component.

Hope that makes sense.

